PR #2, opened 2015-11-20
created mentor and mentee models

KaitCrawford:
Review https://github.com/praekelt/mentorship/pull/1 first

PR #3, opened 2015-11-20
added call model

KaitCrawford:
review and merge https://github.com/praekelt/mentorship/pull/2 first


mentorship/mentor/views.py
```diff
@@ -0,0 +1,54 @@
...
+    })
+    return TemplateResponse(request, 'mentor/login.html', context)
+
+@login_required(login_url='/login/')
+def home(request):
+    user = request.user
+    context = {
+        'user': request.user,
+    }
+    if request.POST:
```
miltontony: Call/End should be 2 separate views. It'll remove the need for the if/else block on `call_state`

mentorship/mentor/views.py
```diff
@@ -0,0 +1,54 @@
...
+        if request.POST['call_state']:
+            call_state = request.POST['call_state']
+            mentor_number = user.mentor_profile.contact_number
+            mentee_number = user.mentor_profile.mentee.contact_number
+            
+            if call_state == 'start':
+                call = Call.objects.create(caller_number=mentor_number, receiver_number=mentee_number)
```
miltontony: With the refactor we chatted about last week, you'll only need to provide `request.user` as the mentor. To initiate the call, you'd do something like `call.dial()` The logic for the numbers to use will be in the dial method.

mentorship/mentor/views.py
```diff
@@ -0,0 +1,54 @@
...
+                context.update({
+                    'call_id': call.id
+                })
+            elif call_state == 'end':
+                call_id = int(request.POST['call_id'])
+                call = Call.objects.get(pk=call_id)
+                call.end_time=datetime.datetime.now()
+                call.save()
+                
+    return TemplateResponse(request, 'mentor/home.html', context)
```
miltontony: The recommended way to render a template is to use [render](https://docs.djangoproject.com/en/1.8/topics/http/shortcuts/#render) `from django.shortcuts import render`. It takes the same arguments as TemplateResponse

mentorship/mentor/views.py
```diff
@@ -0,0 +1,54 @@
...
+            mentee_number = user.mentor_profile.mentee.contact_number
+            
+            if call_state == 'start':
+                call = Call.objects.create(caller_number=mentor_number, receiver_number=mentee_number)
+                context.update({
+                    'call_id': call.id
+                })
+            elif call_state == 'end':
+                call_id = int(request.POST['call_id'])
+                call = Call.objects.get(pk=call_id)
```
miltontony: It's usually safer to use [get_object_or_404](https://docs.djangoproject.com/en/1.8/topics/http/shortcuts/#get-object-or-404). It has better error handling for cases that the object you're looking for doesn't exist.

KaitCrawford:
@miltontony ready for review. please note that due to the model refactor I deleted the migrations and haven't added new ones yet

miltontony:
:+1: :cake:  Looks soo much better!

PR #10, opened 2015-11-27
Api models and call feature

KaitCrawford:


mentorship/mentor/models.py
```diff
@@ -32,7 +33,8 @@ class Call(models.Model):
     receiver = models.ForeignKey('Contact', related_name="receiver")
 
     def dial(self):
-        pass
+        api = comms_models.Api.objects.last()
+        api.make_call(self.caller.contact_number, self.receiver.contact_number)
```
KaitCrawford: @miltontony @smn I need help here please. I can't figure out how to get this to call the subclasses make_call() rather than the superclasses make_call() without specifying the subclass which I don't want to do (I want it to be able to use any subclass)

hodgestar: Why do you need to call the subclass' version of `make_call`? (apologies for the drive-by PR reading)

KaitCrawford:
@miltontony ready for review

miltontony:
:+1: 
I didn't know about the new celery config stuff... you learn something new everyday..

mentorship/mentor/models.py
KaitCrawford: @hodgestar it's because we don't want this code to care what subclass of Api is being used. (All subclasses of Api should define their own make_call() method)

PR #13, opened 2015-12-14
API scaffold and implementation of call_end api endpoint

KaitCrawford:


KaitCrawford:
@miltontony ready for review. Does Not require previous PRs :smile:

miltontony:
:cake: this looks great. Can you add a few tests so that we can see it in action?

miltontony:
:+1:

PR #18, opened 2016-01-05
Small fix for the latest version of django-polymorphic

KaitCrawford:


KaitCrawford:
@miltontony @rudigiesler small review. A new version of django-polymorphic was released at the end of december and it broke the import. This is a fix for that

hodgestar:
:+1:

PR #23, opened 2016-01-12
ReadTheDocs documentation

KaitCrawford:


hodgestar:
Looks good!

Has this been set up in readthedocs.org yet? Can we set it up, get readthedocs.org to build the docs from this branch and post a link to the built docs here? That should allow us to check that the docs build correctly.

KaitCrawford:
@hodgestar finally got it working at http://mentorship-tool.readthedocs.org/en/latest/

KaitCrawford:
@miltontony when you have the time would you run through these and make sure you agree with the content? thanks

hodgestar:
:+1: from me on landing this -- we can always make changes in later PRs if needed.

docs/overview.rst
```diff
@@ -0,0 +1,24 @@
+Moving Parts
+============
+The Mentoring Tool has a number of moving parts described below::
+
+
+1. Comms API:
+    REST API to handle communications to and from messaging and voice call
```
hodgestar: One minor comment -- I think that if you format this as:
```
1. Comms API

   REST API ...

   * Receiver ...
   * Receiver ...
```
it might be rendered more prettily by Sphinx.

hodgestar:
And do we maybe want to say something about how authentication with the API works?

docs/overview.rst
KaitCrawford: I've made each numbered section a header instead

KaitCrawford:
@hodgestar added the requested information about how to authenticate with the API

KaitCrawford:
@rudigiesler @saeedmb8 just tagging you for your interest.
The link to these docs is in the conversation.

PR #31, opened 2016-02-09
Call API endpoint for app

KaitCrawford:


mentorship/comms/urls.py
```diff
@@ -1,11 +1,11 @@
...
-from rest_framework.routers import DefaultRouter
+from comms.routers import CommsRouter
 
 # Create a router and register our viewsets with it.
-router = DefaultRouter()
-router.register(r'call', views.CallViewSet)
-router.register(r'message', views.MessageViewSet)
+router = CommsRouter()
+router.register(r'call', views.CallViewSet, "comms-call")
+router.register(r'message', views.MessageViewSet, "message")
```
hodgestar: Should we perhaps name this `comms-message` and give names to the other `CommsRouter` views for consistency?

KaitCrawford: I was planning to do that in the PRs where I deal with those sections. so the message PR would add names to the message view, etc.

KaitCrawford:
@hodgestar Now it's ready for review :)
@miltontony @saeedmb8 just to keep you in the loop

mentorship/comms/tests/test_comms_views.py
```diff
@@ -149,18 +157,24 @@ def test_message_detail_get(self):
...
+
+        # test that login is required
+        response = self.client.post(
+            reverse('message-list'), {
+                'sender_contact': mentee.contact_number,
+                'recipient_contact': mentor.contact_number,
+                'content': "a message!"
+            })
+        self.assertEqual(response.status_code, 403)
+        self.client.login(username='tester', password='tester')
```
hodgestar: Should the login-required test perhaps be in a separate test?

mentorship/comms/tests/test_comms_views.py
```diff
@@ -208,14 +223,20 @@ def test_feedback_detail_get(self):
...
+        call = self.mk_call()
+
         # test that login is required
-        response = self.client.get(reverse('call-list'))
+        response = self.client.post(
+            reverse('feedback-list'), {
+                'sender_id': call.receiver.pk,
+                'call_id': call.pk,
+                'content': "{'this': 'is','an': {'ivr': 'feedback',},}"
+            })
```
hodgestar: And this one too?

hodgestar:
I've sometimes found it quite useful to write small helper functions like:
```
def check_api_call_successful(self, route, expected_json):
      ....
```
and then tests become:
```
def test_foo(self):
      ... do some setup ...
      self.check_api_call_successful(reverse(...), {
           ... json stuff ...
      })
```

mentorship/mentor/views.py
```diff
@@ -10,6 +10,123 @@
...
+
+
+class CallViewSet(mixins.ListModelMixin, mixins.RetrieveModelMixin,
+                  mixins.CreateModelMixin, mixins.UpdateModelMixin,
+                  viewsets.GenericViewSet):
+    """
+    This viewset automatically provides `list`, `create`, `detail` and `update`
+    actions.
+    """
+    queryset = Call.objects.all()
```
hodgestar: Should we have `queryset` here given that we also have `get_queryset`?

mentorship/mentor/views.py
```diff
@@ -10,6 +10,123 @@
...
+    serializer_class = CallSerializer
+    permission_classes = (permissions.IsAuthenticated,)
+
+    def get_queryset(self):
+        current_user = self.request.user.contact_details
```
hodgestar: Perhaps this should be called `contact_details` and not `current_user` (since that's presumably `request.user`)?

mentorship/mentor/views.py
```diff
@@ -10,6 +10,123 @@
...
+        return Call.objects.filter(
+            Q(caller=current_user) | Q(receiver=current_user))
+
+    def perform_create(self, serializer):
+        user = self.request.user
```
hodgestar: We don't really use `user` -- we only use `contact_details`?

mentorship/mentor/views.py
```diff
@@ -10,6 +10,123 @@
...
+    # Sets the end_time for a call to the current time
+    def perform_update(self, serializer):
+        call = self.get_object()
+
+        if call.end_time is not None:
+            raise ValidationError('This call has already ended.')
+
+        # The change in this line will be overwritten, it is here to ensure
+        # that the response displays the new data.
+        serializer.save(end_time=datetime.datetime.now())
```
hodgestar: Should we get the correct end time from whatever writes the end time into the database model? Maybe after calling `call.ended()`?

mentorship/mentor/views.py
```diff
@@ -10,6 +10,123 @@
...
+
+        mentor = call.caller.mentor
+        call.ended()
+        Event.objects.create(description="Call ended",
+                             event_type="end_call",
+                             relevant_mentor=mentor)
```
hodgestar: I'm wondering a bit about creating the `Event` objects here and in create. It feels a bit like it would be better if these were created by `call.dial()` and `call.ended()`?

mentorship/mentor/views.py
```diff
@@ -10,6 +10,123 @@
...
+
+    def get_queryset(self):
+        current_user = self.request.user.contact_details
+        return Message.objects.filter(
+            Q(sender=current_user) | Q(recipient=current_user))
+
+    def perform_create(self, serializer):
+        user = self.request.user
+        sender = user.contact_details
+        recipient = sender.mentor.mentee
```
hodgestar: Not for this PR, but requiring that there be a one-to-one mapping between mentors and mentees seems a bit restrictive.

mentorship/mentor/views.py
```diff
@@ -10,6 +10,123 @@
...
+        data = serializer.validated_data
+        serializer.save(sender=sender, recipient=recipient)
+
+        Event.objects.create(
```
hodgestar: Hmm. What actually sends the message? Should that create the event?

mentorship/mentor/views.py
```diff
@@ -10,6 +10,123 @@
...
+            event_type="send_message",
+            relevant_mentor=sender.mentor)
+
+
+class EventViewSet(mixins.ListModelMixin, mixins.RetrieveModelMixin,
+                   viewsets.GenericViewSet):
+    """
+    This viewset automatically provides `list` and `detail` actions.
+    """
+    queryset = Event.objects.all()
```
hodgestar: Same question about needing `queryset` here given that we define `get_queryset`.

mentorship/mentor/views.py
```diff
@@ -10,6 +10,123 @@
...
+        return Event.objects.filter(relevant_mentor=current_user)
+
+
+class TaskViewSet(mixins.ListModelMixin, mixins.RetrieveModelMixin,
+                  mixins.UpdateModelMixin, viewsets.GenericViewSet):
+    """
+    This viewset automatically provides `list`, `detail` and `update`
+    actions.
+    """
+    queryset = Task.objects.all()
```
hodgestar: Same `queryset` question. :)

hodgestar:
Done commenting for now. :)

KaitCrawford:
@hodgestar, @miltontony, @saeed, @trevoroctober ready for re-review

mentorship/mentor/models.py
```diff
@@ -58,13 +64,17 @@ class Message(models.Model):
...
+    def send(self, event_type, event_description, mentor):
         api = comms_models.Api.objects.last()
         api.send_message(
             self.sender.contact_number,
             self.recipient.contact_number,
             self.content
         )
+        Event.objects.create(
+            description=event_description,
+            event_type=event_type,
```
hodgestar: Should the type not always be `send_message` and the description `Sent message: {{content}}`?

mentorship/mentor/serializers.py
```diff
@@ -0,0 +1,37 @@
+from rest_framework import serializers
+from mentor.models import Call, Message, Event, Task, ScheduledCall
+
+
+class CallSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Call
+        fields = ('id', 'start_time', 'end_time', 'caller', 'receiver')
+        read_only_fields = ('caller', 'receiver', 'end_time')
```
hodgestar: Should `start_time` also be read only?

mentorship/mentor/serializers.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class MessageSerializer(serializers.ModelSerializer):
+
+    class Meta:
+        model = Message
+        fields = ('id', 'time_sent', 'sender', 'recipient', 'content')
+        read_only_fields = ('sender', 'recipient')
```
hodgestar: Should `time_sent` and `content` also be read only?

mentorship/mentor/serializers.py
hodgestar: I'm wondering if we shouldn't be making fewer fields read only and adding the read-only permissions at the API permissions level?

mentorship/mentor/tests/test_views.py
```diff
@@ -198,3 +215,401 @@ def test_task_list_view(self):
...
+
+    def check_login_required(self, route, method='get', data={}):
+        if method == 'get':
+            response = self.client.get(route)
+        elif method == 'post':
+            response = self.client.post(route, data)
+        elif method == 'put':
+            response = self.client.put(route, data)
+        elif method == 'delete':
+            response = self.client.delete(route)
```
hodgestar: You can use `response = getattr(self.client, method)(route, **kw)` if you replace `data={}` with `**kw`.

Aside: It's usually a bad idea to put mutual containers as default values because `data={}` only creates one dictionary and not one dictionary each function call (i.e. if someone adds an element to `data`, all future function calls will not see and empy dictionary).

mentorship/mentor/serializers.py
KaitCrawford: `time_sent` doesn't need to be read only. Since this field has `auto_now_add=true` in the model it is automatically treated is read_only.
Content should not be marked as read_only because then you wouldn't be able to set it for new messages. The API Viewset for messages doesn't allow editing of existing messages so this is not a concern.

mentorship/mentor/tests/test_views.py
```diff
@@ -198,3 +215,401 @@ def test_task_list_view(self):
...
+        scheduled_call = ScheduledCall.objects.get(pk=1)
+        self.assertEqual(scheduled_call.caller, self.mentor)
+        self.assertEqual(scheduled_call.callee, self.mentor.mentee)
+        self.assertEqual(scheduled_call.call_time.date(), call_time)
+
+        # check relevant events and messages were created
+        events = Event.objects.all()
+        self.assertEqual(events.count(), 2)
+        messages = Message.objects.all()
+        self.assertEqual(messages.count(), 1)
```
hodgestar: It would probably be better to check that exactly the right events and messages were created, unless this is tested somewhere else?

mentorship/mentor/serializers.py
KaitCrawford: Start_time doesn't need to be read only. Since this field has `auto_now_add=true` in the model it is automatically treated is read_only.
Adding the fields as read_only here allows the instance to be edited but prevents those fields from being changed. (This one is a special case because the thing we need to be able to set, end_time, is a time that is being set automatically)

mentorship/mentor/tests/test_views.py
KaitCrawford: mutual containers? I'm not sure what that is

hodgestar: Oops. *mutable.

KaitCrawford:
@hodgestar ready for re-review. I'll start poking around with the merge conflicts for now

mentorship/mentor/serializers.py
KaitCrawford: After our discussion I decided to leave `end_time` as it is here.
This means that the mentor API will accept an empty PUT request to set `end_time`. I think this is acceptable since this is a developer feature, intended to help them with testing etc. I know we discussed that we'd prefer for them to be able to input an end_time but removing `end_time` from `read_only_fields` makes it editable when creating a new call as well which is not what we want. To prevent this we'd have to make two separate serializers and switch between them based on the request type. Unless you feel like that is necessary I'd rather not add that complication.

hodgestar:
:+1: :tropical_fish:

hodgestar:
I can review again after the merge if you like.

KaitCrawford:
@hodgestar sorry to take up more of your time but yes, I think one more once-over tomorrow morning would probably be for the best

mentorship/comms/tests/test_comms_views.py
```diff
@@ -56,172 +61,184 @@ def mk_msg(self):
             sender=mentee, recipient=mentor, content="new message")
         return msg
 
+    def check_login_required(self, route, method, **kw):
+        response = getattr(self.client, method)(route, **kw)
+        self.assertEqual(response.status_code, 403)
+        self.client.login(username='tester', password='tester')
```
hodgestar: Letting this one slide for now. :)

hodgestar:
:+1:

mentorship/comms/tests/test_comms_views.py
KaitCrawford: :smile: thanks. I'll fix it up later

PR #32, opened 2016-02-09
created mentor and mentee serializer

TrevorOctober:


TrevorOctober:
@hodgestar @KaitCrawford Ready for review

mentorship/mentor/serializers.py
```diff
@@ -0,0 +1,83 @@
...
+    password = serializers.CharField(
+        source='user.password')
+
+    class Meta:
+        model = Mentor
+        fields = ('id', 'username', 'first_name', 'last_name',
+                  'email', 'password', 'contact_number', 'mentee')
+
+    def create(self, validated_data):
+        print validated_data
```
hodgestar: This stray `print` statement should be removed. In general, try to never commit `print` statements. :)

mentorship/mentor/serializers.py
```diff
@@ -0,0 +1,83 @@
...
+from django.contrib.auth.models import User
+
+
+class MentorSerializer(serializers.ModelSerializer):
+    id = serializers.CharField(
+        source='user.id')
+
+    username = serializers.CharField(
+        source='user.username')
+    first_name = serializers.CharField(
```
hodgestar: Missing blank line before this one.

mentorship/mentor/views.py
```diff
@@ -156,3 +160,19 @@ class ScheduledCallDelete(DeleteView):
...
     def dispatch(self, *args, **kwargs):
         return super(ScheduledCallDelete, self).dispatch(*args, **kwargs)
+
+
+class MenteeViewSet(mixins.ListModelMixin, mixins.RetrieveModelMixin,
+                    mixins.CreateModelMixin, mixins.UpdateModelMixin,
+                    viewsets.GenericViewSet):
+    queryset = Mentee.objects.all()
+    serializer_class = MenteeSerializer
+    permission_classes = (permissions.IsAuthenticated,)
```
hodgestar: I don't think anyone who is authenticated should be able to see all mentees or mentors, so I'm guessing we either need a different permission check or to limit the set of mentees and mentors returned.

@KaitCrawford Thoughts?

mentorship/mentor/tests/test_views.py
```diff
@@ -198,3 +200,46 @@ def test_task_list_view(self):
...
+    def mk_subclass_api(self, url='https://example.com'):
+            api = ExampleApi.objects.create(call_url=url, message_url=url)
+            return api
+
+    def mk_user(self):
+        user = User.objects.create(username="ntu2ko",
+                                   first_name="Ntuthuko",
+                                   last_name="Dlamini",
+                                   email="ntu2kotrevor@gmail.com",
+                                   password="1234")
```
hodgestar: We shouldn't use real email addresses (or real-looking email addresses) in tests because they might accidentally get used in various ways. The domain name `example.com` is specifically earmarked for use in these sorts of cases, so we should use `something@example.com` as the mail address.

mentorship/mentor/tests/test_views.py
```diff
@@ -198,3 +200,46 @@ def test_task_list_view(self):
...
+        user.save()
+
+    def test_api_root(self):
+        response = self.client.get(reverse(self.url_name_base + 'api-root'))
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(
+            response, '"mentor":"http://testserver/'
+            + self.app_name + '/mentor/"')
```
hodgestar: It would be much better to parse the response (presumably its JSON, so use `json.loads(...)`) and check the result. That allows us to check that the response is valid JSON and make better assertions about the data returned.

mentorship/mentor/tests/test_views.py
```diff
@@ -198,3 +200,46 @@ def test_task_list_view(self):
...
+
+    def test_create_mentee(self):
+        user = User.objects.create(username="ntu2ko",
+                                   first_name="Ntuthuko",
+                                   last_name="Dlamini",
+                                   email="ntu2kotrevor@gmail.com",
+                                   password="1234")
+
+        mentee = Mentee.objects.create(
+            contact_number="contact_number", user=user)
```
hodgestar: If you're creating the mentee via an API call, you shouldn't be creating it here.

mentorship/mentor/tests/test_views.py
```diff
@@ -198,3 +200,46 @@ def test_task_list_view(self):
...
+
+        response = self.client.post(
+            reverse(self.url_name_base + 'api-root'), {})
```
hodgestar: This is not the URL for creating the a Mentee -- it's the base URL for the API. You'll need to add "/mentee" on the end to get the mentee URL.

mentorship/mentor/tests/test_views.py
```diff
@@ -198,3 +200,46 @@ def test_task_list_view(self):
...
+
+    def test_api_root(self):
+        response = self.client.get(reverse(self.url_name_base + 'api-root'))
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(
+            response, '"mentor":"http://testserver/'
+            + self.app_name + '/mentor/"')
+
+    def test_create_mentee(self):
+        user = User.objects.create(username="ntu2ko",
```
hodgestar: Your tests are failing because you are attempting to create a user with the same name as the one you create in `.setUp()` and users need to have unique names. You don't need to create a user here so you can delete this.

mentorship/mentor/views.py
KaitCrawford: @hodgestar Agreed. Ideally this will never be used by anything other than the MentorTogether code so we'd probably need to create a user or something to provide that with access and then restrict the access here so that nothing else can use it.

hodgestar: @KaitCrawford Maybe we can use http://www.django-rest-framework.org/api-guide/permissions/#djangomodelpermissions?

TrevorOctober:
@hodgestar @KaitCrawford @saeedmb8 @miltontony Ready for Review

mentorship/mentor/tests/test_views.py
```diff
@@ -198,3 +200,106 @@ def test_task_list_view(self):
...
+        self.client.login(username='foo', password='1234')
+
+        response = self.client.get(reverse('mentor-list'))
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, '"count":0')
+
+        self.mk_mentor()
+        response = self.client.get(reverse('mentor-list'))
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, '"count":1')
```
hodgestar: We should check that the response actually contains the expected mentor.

Start by writing a check helper function like the one @KaitCrawford did:
```python
def assert_api_list_response(self, response, data):
    self.assertEqual(json.loads(response.content), {
        'count': len(data),
        'next': None,
        'previous': None,
        'results': data,
    })
```
and then use it 
```python
mentor = self.mk_mentor()
serializer = MentorSerializer(mentor)
...
self.assert_api_list_response(response, [serializer.data])
```
You'll want to replace all the places where you do things like `self.assertContains(response, '"count":0')`.

mentorship/mentor/tests/test_views.py
```diff
@@ -198,3 +200,111 @@ def test_task_list_view(self):
...
+        self.client.login(username='foo', password='1234')
+
+        response = self.client.get(reverse('mentee-list'))
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, '"count":0')
+
+        self.mk_mentee()
+        response = self.client.get(reverse('mentee-list'))
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, '"count":1')
```
hodgestar: We should also user the `assert_api_list_response` helper function here.

mentorship/mentor/tests/test_views.py
```diff
@@ -198,3 +200,111 @@ def test_task_list_view(self):
...
+        self.assertEqual(json.loads(response.content), serializer.data)
+
+    def test_mentee_list(self):
+        response = self.client.get(reverse('mentee-list'))
+        self.assertEqual(response.status_code, 403)
+        self.client.login(username='foo', password='1234')
+
+        response = self.client.get(reverse('mentee-list'))
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, '"count":0')
```
hodgestar: We should also use the `assert_api_list_response` helper here.

TrevorOctober:
@hodgestar @KaitCrawford @miltontony @saeedmb8 Ready for re-review

mentorship/mentor/tests/test_views.py
```diff
@@ -198,3 +200,109 @@ def test_task_list_view(self):
...
+                                   email="foo@example.com")
+        user.set_password('1234')
+        user.is_staff = True
+        user.save()
+        return user
+
+    def mk_mentee(self):
+        mentee = Mentee.objects.create(contact_number="0123456789",
+                                       user=self.user)
+        mentee.save()
```
hodgestar: You can remove `mentee.save()` here because `.create` already saves it.

mentorship/mentor/tests/test_views.py
```diff
@@ -198,3 +200,109 @@ def test_task_list_view(self):
...
+        return mentee
+
+    def mk_mentor(self):
+        mentor = Mentor.objects.create(contact_number="1992011615",
+                                       user=self.user)
+        mentor.save()
```
hodgestar: You can remove `mentor.save()` here because `.create` already saves it.

mentorship/mentor/tests/test_views.py
```diff
@@ -198,3 +200,109 @@ def test_task_list_view(self):
...
+            'results': data,
+        })
+
+    def test_api_root(self):
+        response = self.client.get(reverse(self.url_name_base + 'api-root'))
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(
+            json.dumps(response.data),
+            '{"mentee": "http://testserver/mentor/mentee/", "mentor": "h'
+            'ttp://testserver/mentor/mentor/"}')
```
hodgestar: This should rather be:
```python
self.assertEqual(response.data, {
    "mentee": ...,
    "mentor": ...,
})
```
because dictionary keys are not ordered (so `json.dumps(response.data)` might produce `{"mentor": ..., "mentee"}` instead of `{"mentee": ..., "mentor": ...}`.

TrevorOctober:
@hodgestar Ready for review

hodgestar:
:+1: :cake:

PR #34, opened 2016-04-04
test improvements and add oniondev api

KaitCrawford:
While adding the oniondev api I noticed some things about the existing tests that could be improved


mentorship/comms/models.py
```diff
@@ -91,4 +91,20 @@ def send_message(self, sender_id, recipient_id, content):
             'message': content,
         }
 
-        post_to_api.delay(url, post_data)
+        make_post_request.delay(url, post_data)
+
+
+class OnionDevApi(Api):
+    authId = models.IntegerField()
```
hodgestar: Could we add a docstring to this class and a help string to the authId field?

mentorship/comms/tests/test_comms_models.py
```diff
@@ -100,3 +93,34 @@ def receive_ivr_feedback(self):
...
+class ExampleApiTests(CommsApiTestBase):
+    def mk_example_api(self, url='https://example.com'):
+        api = ExampleApi.objects.create(call_url=url, message_url=url)
+        return api
+
+    def test_example_make_call(self):
+        api = self.mk_example_api()
+        with patch('comms.models.make_post_request.delay') as mock_task:
+            api.make_call('0820000000', '0830000000')
+            self.assertTrue(mock_task.called)
```
hodgestar: We should probably assert that it was called with the correct arguments.

KaitCrawford:
@hodgestar ready for review. I'm trying to do the steps in separate PRs so they are easier to review
cc @TrevorOctober

hodgestar:
:+1:

PR #36, opened 2016-04-05
add dummy authId

TrevorOctober:


mentorship/comms/models.py
```diff
@@ -116,3 +116,14 @@ def make_call(self, caller_id, receiver_id):
         }
 
         make_post_request.delay(url, data)
+
+    def send_message(self, recipient_id, content):
+        url = self.message_url
+
+        data = {
+            'ai': self.schedule_id,
```
KaitCrawford: this needs to be self.auth_id. while schedule_id and auth_id will both be provided by OnionDev we have no reason to believe that they'd be the same.
You can add auth_id as a field on the OnionDevApi model the same way I added schedule_id

mentorship/comms/models.py
```diff
@@ -104,6 +104,11 @@ class OnionDevApi(Api):
         help_text="A unique integer for identifying a request, provided by "
         "OnionDev")
 
+    authId = models.IntegerField(
+        max_length=200,
+        help_text="A unique integer for "
+        "identifying a request, provided by OnionDev")
```
KaitCrawford: Make the help_text a little more specific. Something like "A unique integer for identifying message requests..."

TrevorOctober:

mentorship/comms/models.py
```diff
@@ -91,4 +91,44 @@ def send_message(self, sender_id, recipient_id, content):
...
+    """
+    Class to handle data being passed to and from OnionDev.
+    """
+
+    schedule_id = models.CharField(
+        max_length=200,
+        help_text="A unique integer for identifying a request, provided by "
+        "OnionDev")
+
+    authId = models.IntegerField(
```
hodgestar: I think this field should be called `auth_id` for consistency with our other field names?

hodgestar:
@TrevorOctober Could you merge develop into this branch so that we can see just the changes you've made?

hodgestar:
Once we're close to being done (or possibly in a separate PR) we'll need to add a migration that adds the new fields.

KaitCrawford:
Whoops, good point @hodgestar! I'm going to have to add another field for the feedback surveys so I'll do it at the end of that PR

TrevorOctober:
@hodgestar @KaitCrawford Please review

KaitCrawford:
:+1:

PR #37, opened 2016-04-06
handle call termination notification from oniondev

KaitCrawford:
messy job put together because they're going to be running tests on this tomorrow


KaitCrawford:
@hodgestar @TrevorOctober ready for review

mentorship/comms/models.py
```diff
@@ -24,10 +27,14 @@ def make_call(self):
         to initiate a call with the specific voice call API. """
         raise Exception('Not implemented')
 
-    def receive_call_ended(self, call_id, end_time):
+    def receive_call_ended(self, data):
         """ Default method for handling a call termination notification. Accepts
         the *call_id* and *end_time* for the terminated call and updates its
         end time in the database. """
+        call_id = data['id']
+        end_time = data['end_time']
```
hodgestar: Why did we change the explicit `call_id` and `end_time` parameters to passing through a generic `data` dict?

KaitCrawford: So that the method signature is the same as it is in descendant classes and also so it matches what we call from the new view (CallEndHandler).
I've been thinking about it and I'll probably remove the ability to update the call through the model serializer so it will only be updated from the new view in future.

mentorship/comms/views.py
```diff
@@ -29,6 +32,16 @@ def perform_update(self, serializer):
         serializer.save()
 
 
+class CallEndHandler(APIView):
+    permission_classes = (permissions.IsAuthenticated,)
+
+    def post(self, request, format=None):
+        data = request.data
+        api = Api.objects.last()
+        api.receive_call_ended(data)
```
hodgestar: I think the parsing of `data` and the generation of errors should happen here in the view, rather than in the helper method on the model.

mentorship/comms/models.py
hodgestar: Ah, I see it's to do validation here. See suggestion to do validation in the view.

mentorship/comms/models.py
```diff
@@ -24,10 +27,14 @@ def make_call(self):
...
+        if isinstance(end_time, basestring):
+            end_time = datetime.strptime(end_time[:-6], "%Y-%m-%dT%H:%M:%S.%f")
         from mentor import models as mentor_models
         call = mentor_models.Call.objects.get(pk=call_id)
```
hodgestar: Maybe the `call` object should also be retrieved during validation in the view? Since presumably we need to generate some sort of `4xx` HTTP error if the call doesn't exist or has already ended?

mentorship/comms/models.py
hodgestar: Does anything actually use this base `received_call_ended` method? Maybe the base API should just raise `NotImplemntedError` for its methods and leave implementing them up to the specific API?

mentorship/comms/views.py
KaitCrawford: The problem with that is that the current design is that all APIs will use the same view for but they won't necessarily all expect the same data.
So at this point we don't know what data the api is expecting so we can't validate that it is there and correct.

hodgestar:
@KaitCrawford I feel like I'm not full understanding how the API base is supposed to be used. Maybe you can quickly walk me through how you see it?

hodgestar:
:+1:

The plan is to land this and then to move the `receive_call_ended` logic directly into the `post` handler for it. The Api base class and sub-classes will stay for the outbound API calls (since the base classes and API sub-classes make complete sense when we're calling the outside world).

PR #38, opened 2016-04-07
move receive_call_ended method to CallEndHandler class

TrevorOctober:


mentorship/comms/views.py
```diff
@@ -37,8 +37,13 @@ class CallEndHandler(APIView):
 
     def post(self, request, format=None):
         data = request.data
-        api = Api.objects.last()
-        api.receive_call_ended(data)
+        call_id = data['id']
+        end_time = data['end_time']
+        if isinstance(end_time, basestring):
+            end_time = datetime.strptime(end_time[:-6], "%Y-%m-%dT%H:%M:%S.%f")
+        from mentor import models as mentor_models
```
hodgestar: I think we can import this at the top of the file now (it was in the function earlier to avoid cyclic imports).

mentorship/comms/views.py
```diff
@@ -37,8 +37,13 @@ class CallEndHandler(APIView):
 
     def post(self, request, format=None):
         data = request.data
-        api = Api.objects.last()
-        api.receive_call_ended(data)
+        call_id = data['id']
+        end_time = data['end_time']
+        if isinstance(end_time, basestring):
+            end_time = datetime.strptime(end_time[:-6], "%Y-%m-%dT%H:%M:%S.%f")
```
hodgestar: We should raise a suitable error if `id` or `end_time` are not given or if `end_time` can't be parsed into a datetime.

hodgestar:
Left some comments on how to improve the API method, and we need to remove the old methods, but otherwise this looks perfect.

mentorship/comms/models.py
```diff
@@ -27,18 +27,6 @@ def make_call(self):
         to initiate a call with the specific voice call API. """
         raise Exception('Not implemented')
 
-    def receive_call_ended(self, data):
```
KaitCrawford: @hodgestar shouldn't this method stay here but raise a "Not implemented" exception?

mentorship/comms/serializers.py
```diff
@@ -10,10 +10,8 @@ class Meta:
         read_only_fields = ('caller', 'receiver')
 
     def save(self):
```
KaitCrawford: @hodgestar I think this method should be removed and the view that uses it be changed to not accept updates.

KaitCrawford:
Shouldn't this PR also include a new view for the onionDev receive_call_ended() method?

mentorship/comms/serializers.py
```diff
@@ -10,10 +10,8 @@ class Meta:
...
-        api = Api.objects.last()
-        data = self.data.copy()
-        data.update(self.validated_data)
-        api.receive_call_ended(data)
+        call = mentor_models.Call.objects.get(pk=self.validated_data['id'])
```
KaitCrawford: @TrevorOctober the tests are failing here because you there isn't an id being passed through in the data. However, in model serializers you don't need the id to get the object you are trying to work with. Try using self.instance to get the call.

mentorship/comms/models.py
hodgestar: My plan was not to have the incoming methods on the API model at all (i.e. receiving HTTP requests and making HTTP requests are separated entirely).

mentorship/comms/serializers.py
hodgestar: This is the API that was for testing / debugging, right? Do we not want to leave it in place? Happy for it to be removed if you prefer.

hodgestar:
@KaitCrawford Re OnionDev view -- I thought that's what https://github.com/praekelt/mentorship/pull/38/files#diff-5b3af93a7af06c0d613fb7ab42a1e832R38 was? If not, we probably have some fixing to do.

mentorship/comms/serializers.py
KaitCrawford: yes, but the new view replaces it

KaitCrawford:
Ah @hodgestar No, that view is a replacement for the receive_call_ended that was in the base api class (the one that was used for testing / debugging). We'll need a new view for the OnionDev version of that method

TrevorOctober:

KaitCrawford:
:+1:

PR #39, opened 2016-04-11
move receive_call_ended method from OnionDevApi model

TrevorOctober:


mentorship/comms/views.py
```diff
@@ -46,6 +47,30 @@ def post(self, request, format=None):
         return HttpResponse(status=200)
 
 
+class OnionDevApiCallEndHandler(APIView):
+    permission_classes = (permissions.IsAuthenticated,)
+
+    def post(self, request, format=None):
+        data = request.data
+        """ Method for handling call results sent when a call has ended """
+        caller_id = mentor_models.Contact.objects.get(
```
KaitCrawford: maybe add a try block here and if we don't get a number for an existing mentor throw a validation error

mentorship/comms/views.py
```diff
@@ -46,6 +47,30 @@ def post(self, request, format=None):
...
+            contact_number=data['handler'])
+        end_time = timezone.make_aware(
+            datetime.strptime(
+                data['conversationEndTime'], "%Y-%m-%d %H:%M:%S.%f"),
+            timezone.get_default_timezone())
```
KaitCrawford: maybe add try blocks or some sort of check around these assignments that throw a validation error if we don't have a mentor with that number or if the time is in the wrong format

KaitCrawford:
@TrevorOctober please add an entry to the urls file and a test for the new view

TrevorOctober:
@KaitCrawford Ready for review

mentorship/comms/urls.py
```diff
@@ -12,6 +12,8 @@
 # Include the login URLs for the browsable API.
 urlpatterns = [
     url(r'^call_end/$', views.CallEndHandler.as_view(), name='comms-call-end'),
+    url(r'^call_end/$', views.OnionDevApiCallEndHandler.as_view(),
```
KaitCrawford: this needs to be something like `r'^od_call_end/$'` or `r'^onion_dev_call_end/$'` or something. otherwise any request for a url that matches will match both patterns and only the first pattern will be used

mentorship/comms/tests/test_comms_views.py
```diff
@@ -147,6 +147,22 @@ def test_call_end_handler(self):
         call = mentor_models.Call.objects.get(pk=call.pk)
         self.assertEqual(call.end_time, end_time)
 
+    def test_call_end_handler_onion_dev(self):
+        url = reverse(self.url_name_base + 'call-end-onion-dev')
+        end_time = timezone.now() + datetime.timedelta(hours=1)
+        data = {"end_time": end_time.isoformat(), "id": 1}
```
KaitCrawford: This dictionary needs to contain the data that the new view expects

KaitCrawford:
:+1:

PR #40, opened 2016-04-12
OnionDev IVR feedback view

KaitCrawford:


KaitCrawford:
Requires #39 to land

mentorship/comms/views.py
```diff
@@ -77,6 +79,36 @@ def post(self, request, format=None):
         return HttpResponse(status=200)
 
 
+class OnionDevApiFeedbackHandler(APIView):
+    permission_classes = (permissions.IsAuthenticated,)
+
+    def post(self, request, format=None):
+        data = request.data
+        """ Method for handling feedback results"""
```
hodgestar: The docstring needs to be directly after the function definition (so above `data = request.data`), otherwise it's just a string.

mentorship/comms/views.py
```diff
@@ -77,6 +79,36 @@ def post(self, request, format=None):
...
+        except ObjectDoesNotExist:
+            raise ValidationError(
+                "The given callerid does not exist."
+            )
+
+        call = mentor_models.Call.objects.filter(receiver=sender).last()
+        if call.end_time is None:
+            raise ValidationError('This call has not ended.')
+
+        str_content = json.dumps(content)
```
hodgestar: I'm not quite sure why we're converting this to JSON here -- shouldn't it just be a string (which it is already)?

hodgestar: After discussion IRL, the content is not a string but rather an object of some sort.

mentorship/comms/tests/test_comms_views.py
```diff
@@ -241,20 +241,19 @@ def test_feedback_detail_get(self):
...
         call = self.mk_call()
-        url = reverse('feedback-list')
+        call.ended()
+        url = reverse(self.url_name_base + 'feedback-onion-dev')
         data = {
-            'sender_id': call.receiver.pk,
-            'call_id': call.pk,
-            'content': "{'this': 'is','an': {'ivr': 'feedback',},}"
+            "callerid": "1234",
+            "recordId": "{'this': 'is','an': {'ivr': 'feedback',},}"
```
hodgestar: If we're meant to be passed structured data, should this be:
```json
"recordId": {
  "this": "is",
  "an": {
     "ivr": "feedback",
  }
}
```
i.e. structured data rather than a string.

mentorship/comms/tests/test_comms_views.py
```diff
@@ -241,20 +241,19 @@ def test_feedback_detail_get(self):
...
         }
-
         # test that login is required
         self.check_login_required(route=url, method="post", data=data)
 
         # test creating new feedback
-        self.check_api_post_successful(url, data)
+        response = self.client.post(url, data)
```
hodgestar: I have a feeling we're not posting and decoding the data to and from JSON (I'd expect us to have to encode the data to JSON explicitly, but maybe it happens magically).

hodgestar:
:+1:

PR #41, opened 2016-04-12
add funtionality to task stubs

KaitCrawford:


mentorship/comms/tasks.py
```diff
@@ -1,15 +1,28 @@
...
+    requests.post(
+        url=url,
+        data=json.dumps(data),
+        headers={
+            'Content-Type': 'application/json',
+        }
+    )
     print "we still need to implement the api call!"
     print data
     return "success"
```
hodgestar: We can remove the prints, and we should use `raise_for_status` to raise an error if there is a non-success response code (http://docs.python-requests.org/en/master/api/#requests.Response.raise_for_status).

hodgestar:
:+1:

PR #45, opened 2016-04-12
added support for authorization header to tasks

KaitCrawford:


KaitCrawford:
@TrevorOctober @hodgestar ready for review

mentorship/comms/models.py
```diff
@@ -88,6 +88,18 @@ class OnionDevApi(Api):
...
+        max_length=100,
+        help_text="The username used to authenticate with their server",
+        default=""
+    )
+
+    password = models.CharField(
+        max_length=200,
+        help_text="The password used to authenticate with their server",
+        default=""
+    )
```
hodgestar: The post and get request methods check whether `password` and `username` are `None`, but that would require these fields to be nullable and probably to default to `None` rather than `""`?

mentorship/comms/tests/test_comms_tasks.py
```diff
@@ -15,11 +15,31 @@ def test_make_post_request(self):
         self.assertEqual(result, "success")
 
     @responses.activate
-    def test_make_get_request(self):
+    def test_make_post_request_auth(self):
+        responses.add(responses.POST, "http://example.com",
+                      body='{"data": "data"}', status=200,
+                      content_type='application/json')
```
hodgestar: Ooh, we can pass `json={"data": "data}` instead of `body` and `content_type`.

mentorship/comms/tests/test_comms_tasks.py
```diff
@@ -15,11 +15,31 @@ def test_make_post_request(self):
...
+        result = make_post_request(
+            "http://example.com", {'data': 'data'}, "username", "password")
+        assert len(responses.calls) == 1
```
hodgestar: We should assert that the username and password were actually sent. Possibly we can do this with:
```
[(request, response)] = responses.calls
self.assertTrue('Authorization' in request.headers)
```
or something similar.

mentorship/comms/tasks.py
```diff
@@ -4,26 +4,32 @@
...
 
 
 @app.task
-def make_get_request(url, data):
+def make_get_request(url, data, username=None, password=None):
+    headers = {
+        'Content-Type': 'application/json'
+    }
+    if username is not None and password is not None:
+        headers['Authorization'] = username + ':' + password
```
hodgestar: We need to b64encode this and add a `Basic: ` in front.

hodgestar:
Done with initial review.

KaitCrawford:
@hodgestar, @TrevorOctober ready for rereview. If you're reviewing per commit note that there were two commits before your last message.

PR #48, opened 2016-04-19
Bring our system up to date with the newest docs

KaitCrawford:


mentorship/comms/views.py
```diff
@@ -56,28 +56,32 @@ def post(self, request, format=None):
...
-        try:
-            end_time = timezone.make_aware(
-                datetime.strptime(
-                    data['conversationEndTime'], "%Y-%m-%d %H:%M:%S.%f"),
-                timezone.get_default_timezone())
-        except KeyError:
-            raise ValidationError("Request must contain 'conversationEndTime'")
+        end_time = timezone.make_aware(
+            datetime.strptime(
+                data['conversationEndTime'], "%Y-%m-%d %H:%M:%S.%f"),
```
hodgestar: Two things:
* We should use `end_time` here instead of `data['conversationEndTime']`since we went to the trouble of assigning it earlier.
* `datetime.strptime` can fail with a parsing error if the type of the value is wrong or the string isn't in the expected format. We should catch that and raise a `ValidationError`.
* Is `timezone.get_default_timezone()` really sane? I'm hoping that the times they send us are in UTC?

mentorship/comms/views.py
```diff
@@ -88,8 +92,8 @@ def post(self, request, format=None):
         """ Method for handling feedback results"""
         data = request.data
         try:
-            caller_id = data['callerid']
-            content = data['recordId']
+            caller_id = data['cdr']['callerid']
+            content = data['records']
         except KeyError:
```
hodgestar: If `data['cdr']` is, e.g. a string, `data['cdr']['x']` could raise an error that is not a `KeyError`. We should handle that case and throw a ValidationError if it's not a dictionary.

hodgestar:
Thank you for working on all these post-original-docs changes!

mentorship/comms/views.py
KaitCrawford: With respect to your last point: Honestly I don't know. The timezone-aware and timezone-naive datetimes confuse and frustrate the bananas out of me but this worked so I stuck with it.

KaitCrawford: I'll raise it in the call with them tomorrow and we can address it once we're sure what timezone they'll be sending us the data in

KaitCrawford:
@hodgestar ready for re-review. I've addressed all of your concerns except the timezone one which I will handle after today's call

hodgestar:
We should add the `sender_id`, `recipient_id and `content` parameters to the base `API.send_message(...)` function definition too.

KaitCrawford:
@hodgestar are you happy for me to leave that for now so we can land this and do it in the PR with the timezone?

mentorship/comms/views.py
```diff
@@ -56,28 +56,36 @@ def post(self, request, format=None):
...
             end_time = timezone.make_aware(
-                datetime.strptime(
-                    data['conversationEndTime'], "%Y-%m-%d %H:%M:%S.%f"),
+                datetime.strptime(end_time, "%Y-%m-%d %H:%M:%S.%f"),
                 timezone.get_default_timezone())
-        except KeyError:
-            raise ValidationError("Request must contain 'conversationEndTime'")
+        except ValueError:
+            raise ValidationError(
+                "'conversationEndTime' Is in the incorrect format."
```
hodgestar: s/Is/is/

hodgestar:
Two very minor comments, otherwise looks good.

hodgestar:
:+1:

PR #49, opened 2016-04-20
Receive message endpoint for onion dev

KaitCrawford:


KaitCrawford:
@hodgestar, @TrevorOctober ready for review in the morning

mentorship/comms/views.py
```diff
@@ -122,10 +122,41 @@ def post(self, request, format=None):
...
+            )
+
+        recipient = mentor_models.Mentor.objects.get(mentee=sender)
+        mentor_models.Message.objects.create(sender=sender,
+                                             recipient=recipient,
+                                             content=content)
+        mentor_models.Event.objects.create(
+            description="Received message: '" + content + "'",
+            event_type="receive_message",
+            relevant_mentor=recipient.mentor)
```
hodgestar: If a mentor can have more than one mentee (not sure if this is allowed) how will they know which mentee this message is from?

hodgestar:
One question, otherwise look good to me.

mentorship/comms/views.py
KaitCrawford: I'm reasonably sure mentor's can only have one mentee but I'll check. Either way, we store the sender with the message so we can display it however we like.

KaitCrawford: Confirmed, the relationship will always be one-to-one

hodgestar:
:+1:

PR #54, opened 2016-05-04
Extra logging for feedback response

KaitCrawford:


KaitCrawford:
@miltontony please review

mentorship/comms/views.py
```diff
@@ -123,10 +123,11 @@ def post(self, request, format=None):
             content = data['records']
         except KeyError:
             logging.error(
-                "Both 'callerid' and 'recordId' fields must be present.",
+                "Both 'callerid' and 'records' fields must be present." +
+                json.dumps(data['cdr']),
```
hodgestar: I think perhaps we need to indicate what this data is, e.g. `"... present (cdr: %s)." % json.dumps(data['cdr']) `.

Also, what is `data['cdr']`? Is it guaranteed to exist? I would have thought it might be more useful to log all of `data`?

miltontony: sentry provides data as a context in the view, so no need to log it separately.

KaitCrawford: @hodgestar The idea behind this change is that I'm not sure OnionDev are currently sending the callerid in the way we expect it so I'm trying to see how they are sending it. Since it's supposed to be in the 'cdr' field I've decided to add that to our logs.

KaitCrawford:
ready for re-review

hodgestar:
@KaitCrawford I still think we should dump out everything since otherwise we'll need to modify the code again to look at something else?

KaitCrawford:
@hodgestar @miltontony ready for re-review

hodgestar:
:+1:

PR #57, opened 2016-05-10
moved LOGGING settings to test a theory

KaitCrawford:


KaitCrawford:
:+1: from @miltontony

PR #60, opened 2016-05-24
Create endpoints for password reset from app

KaitCrawford:


KaitCrawford:
@hodgestar @miltontony @rudigiesler ready for review

mentorship/mentor/views.py
```diff
@@ -180,6 +189,60 @@ def perform_destroy(self, serializer):
...
+            return response.Response(data=serializer.errors,
+                                     status=status.HTTP_400_BAD_REQUEST)
+
+        data = serializer.validated_data
+        token = data['token']
+        requests = ResetPasswordRequest.objects.filter(
+            token=token)
+        if requests.count() <= 0:
+            return response.Response(status=status.HTTP_401_UNAUTHORIZED)
+        user = requests.latest('created_at').user
```
hodgestar: I think we need to have some concept of a password reset request expiring. Perhaps if it is older than a certain time?

hodgestar: The one down side to relying purely on the time is that temporary system clock failures can then be a security hole, which is a bit scary. Alternatively we could have an expired / not-expired field on the model, but then what would set that?

KaitCrawford: default_token_generator handles this for us. If we set PASSWORD_RESET_TIMEOUT_DAYS in settings.py then it will expire the token after the given number of days.
You can see the code here if you're interested https://github.com/django/django/blob/master/django/contrib/auth/tokens.py#L42

hodgestar: Woot! :)

KaitCrawford:
@hodgestar ready for re-review.
Locally i've added `EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'` to settings.py for testing. Do we need to add this setting and give it a logical value for production or is the default fine?

hodgestar:
Looks good! Let's add the migrations.

Regarding production -- yes, there are some settings we should check, but we can discuss that in internal communication channels.

hodgestar:
:+1:

PR #62, opened 2016-05-26
changed the 'from' email address

KaitCrawford:


mentorship/mentor/utils.py
```diff
@@ -1,10 +1,9 @@
-from django.conf import settings
 from django.core.mail import EmailMultiAlternatives
 from django.template import loader
 
 
 def send_reset_email(to_email, context):
-    from_email = settings.DEFAULT_FROM_EMAIL
+    from_email = "no-reply@mentortogether.org"
```
hodgestar: I sort of feel we should either just set `DEFAULT_FROM_EMAIL` to this in our config settings, or create another setting for this if it needs to be different to `DEFAULT_FROM_EMAIL`.

KaitCrawford:
@hodgestar please review

PR #65, opened 2016-06-01
added new fields

TrevorOctober:
@KaitCrawford Please review


dump.rdb
```diff
@@ -0,0 +1 @@
+REDIS0006�ܳC�Z��V
```
KaitCrawford: Please don't commit this file. Delete it or add it to .gitignore or something

mentorship/mentor/models.py
```diff
@@ -24,6 +24,12 @@ class Mentee(Contact):
 class Mentor(Contact):
     mentee = models.OneToOneField(Mentee, related_name="mentee", null=True,
                                   blank=True)
+    profile_pic = models.ImageField(upload_to="media/", null=True, blank=True)
+    job_sector = models.CharField(max_length=30, null=True, blank=True)
+    job_title = models.CharField(max_length=30, null=True, blank=True)
+    motivation = models.CharField(max_length=200, null=True, blank=True)
+    inspiration = models.CharField(max_length=200, null=True, blank=True)
+    tags = models.CharField(max_length=200, null=True, blank=True)
```
KaitCrawford: My gut feeling is that you should probably increase the max_length on these. Make the shorter ones about 100 and the longer ones you can either change to TextField or make the max_length something like 500

TrevorOctober:
@KaitCrawford Ready for re-review

KaitCrawford:
:+1: Looks good

PR #66, opened 2016-06-02
Push notifications and settings

KaitCrawford:


KaitCrawford:
Needs test

mentorship/mentor/models.py
```diff
@@ -155,3 +172,28 @@ class ResetPasswordRequest(models.Model):
...
+class NotificationSettings(models.Model):
+    mentor = models.OneToOneField('Mentor',
+                                  related_name="notification_settings")
+    android_registration_token = models.CharField(
+        max_length=200, null=True,
+        help_text="Token used to send push notifications")
+    ten_min_call_reminder = models.BooleanField(default=True)
+    one_hour_call_reminder = models.BooleanField(default=True)
+    new_message = models.BooleanField(default=True)
+    new_tip = models.BooleanField(default=True)
```
hodgestar: Could we add help_text for some of the other fields too (e.g. I'm not sure what `new_message` or `new_tip` do).

mentorship/mentor/serializers.py
```diff
@@ -76,12 +76,13 @@ def create(self, validated_data):
             contact_number=validated_data['contact_number'],
             user=user)
         mentor.save()
+        NotificationSettings.objects.create(mentor=mentor)
```
hodgestar: I wonder if this should rather be a post-create hook on mentors since otherwise mentors can be created without notification settings (e.g. via the admin)? Thoughts?

mentorship/mentor/serializers.py
```diff
@@ -128,3 +129,12 @@ class NewPasswordSerializer(serializers.Serializer):
...
         style={'input_type': 'password'})
+
+
+class NotificationSettingsSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = NotificationSettings
+        fields = ('mentor', 'android_registration_token',
+                  'ten_min_call_reminder', 'one_hour_call_reminder',
+                  'new_message', 'new_tip')
+        read_only_fields = ('mentor')
```
hodgestar: I think you need a `,` after `'mentor'` because otherwise this is just a bracketed string rather than a tuple with one element.

mentorship/mentor/tasks.py
```diff
@@ -0,0 +1,30 @@
+import datetime
+from mentor.models import ScheduledCall
+from mentorship.celery import app
+
+
+@app.task
+def send_upcoming_calls_reminder():
+    current_minute = datetime.datetime.now().replace(second=0, microsecond=0)
```
hodgestar: Is the idea that this task will be run every minute? I'm a bit concerned that that will create a problem where a task might be run at `10:00:00.1`, `10:00:59.9` and `10:02:00.1`` and thus `10:00` would be processed twice and `10:01` not at all.

hodgestar: Probably the only way to avoid this is to get all calls that are less than or equal to 10 minutes away and mark those which have had reminders sent and not process those in future runs.

Maybe there is a smarter way, but I'm not thinking of it right now.

hodgestar:
I've left a couple of comments, but otherwise things look good to me so far.

KaitCrawford:
@hodgestar ready for review :smile:

mentorship/mentor/tests/test_models.py
```diff
@@ -0,0 +1,235 @@
...
+        mentor_user = self.mk_user(name)
+        mentor = Mentor.objects.create(
+            user=mentor_user,
+            contact_number='1234')
+        if mentee:
+            mentor.mentee = mentee
+        mentor.save()
+        return mentor
+
+    def test_send_calls_reminder(self):
```
hodgestar: As a general rule, I would break tests like this up into a series of smaller test cases that each test one thing (and write lots of helper methods so code isn't duplicated between tests). This is more of a general thing though and not necessarily something that needs to be done for this PR.

mentorship/mentor/tests/test_models.py
```diff
@@ -0,0 +1,235 @@
...
+        ScheduledCall.objects.create(
+            caller=mentor, callee=mentee, call_time=ten_min_from_now)
+        with patch.object(ScheduledCall, 'remind_caller', return_value=True)\
+                as mock_method:
+            result = send_upcoming_calls_reminder()
+            mock_method.assert_called_once_with()
+        self.assertEqual(result, "success")
+        reminders_sent = CallReminder.objects.all()
+        self.assertEqual(len(reminders_sent), 2)
+        self.assertEqual(reminders_sent[1].reminder_type, "ten_min_reminder")
```
hodgestar: Do we also need tests that once a call reminder has been sent, further calls to `send_upcoming_calls_reminder` don't cause the reminder to be sent again?

mentorship/mentor/tests/test_views.py
```diff
@@ -301,6 +301,7 @@ def check_api_post_successful(self, route, data):
 
     def check_api_put_successful(self, route, data):
         response = self.client.put(route, data, 'application/json')
+        print response
```
hodgestar: Stray debugging print.

mentorship/mentorship/settings.py
```diff
@@ -11,9 +11,15 @@
 """
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
+from __future__ import absolute_import
```
hodgestar: Is there a particular change that required this? Absolute imports are generally a good idea, just curious.

hodgestar:
Left a few questions but otherwise looks good.

mentorship/mentorship/settings.py
KaitCrawford: I think it was for importing crontab in settings.py but it might have been for importing djcelery

mentorship/mentor/tests/test_models.py
KaitCrawford: yes

KaitCrawford:
@hodgestar ready for re-review. I also added migrations

hodgestar:
:+1:

PR #71, opened 2016-06-06
Better error handling for mentor and mentee API endpoints

KaitCrawford:


mentorship/mentor/serializers.py
```diff
@@ -58,11 +58,20 @@ class MentorSerializer(serializers.ModelSerializer):
...
+    contact_number = serializers.CharField(required=True)
+
     class Meta:
         model = Mentor
         fields = ('id', 'username', 'first_name', 'last_name',
                   'email', 'password', 'contact_number')
 
+    def validate_username(self, value):
+        conflicting_users = User.objects.filter(username=value)
+        if conflicting_users.count() > 0:
```
hodgestar: Will this prevent submitting the `username` when updating a mentor's other details?

KaitCrawford: Hmmm good point. That also brings up another good question: should users be able to change their username, first_name, last_name, email and contact_number?

KaitCrawford: I don't think this serializer currently handles that

hodgestar: So we need to add a custom update method too?

KaitCrawford: I'm inclined to say no at present. User accounts must have admin status in order to use this API endpoint because it is intended to be called by the Mentor Together system to create new Mentors. There has been no requirement that updates on that system should be pulled through.
That said, I will check with the PM and also confirm if these should be up-datable from the app. That should inform what we do here.

hodgestar:
I'm :+1: on landing this and then creating an issue to create a custom update method if needed. Tests for that should pick up the validation issue, and the validation is correct for Mentee/or creation.

KaitCrawford:
@hodgestar Update after discussion: Update feature is not required for these fields

hodgestar:
Woot.

:+1:

PR #81, opened 2016-07-01
Return the profile pic URL

hodgestar:


hodgestar:
Ready for review.

mentorship/mentor/serializers.py
```diff
@@ -133,12 +134,30 @@ def create(self, validated_data):
...
+        return url
+
+
 class ProfileSerializer(serializers.ModelSerializer):
 
-        class Meta:
-            model = Mentor
-            fields = ('job_sector', 'job_title',
-                      'motivation', 'inspiration', 'tags')
+    profile_pic = ProfilePicUrlField(required=False, read_only=True)
```
KaitCrawford: Can't we use Django rest frameworks HyperlinedModelSerializer in some way to simplify this stuff? http://www.django-rest-framework.org/api-guide/serializers/#hyperlinkedmodelserializer

KaitCrawford:
@hodgestar just one question about Hyperlinked Serializers.

mentorship/mentor/serializers.py
hodgestar: There doesn't seem to be a way to get a URL for a detail route on a view (i.e. adding the `image/` on the end). We could override `HyperlinkedIdentityField` instead of `Field`. We'd still have to override `to_representation`, but maybe it'd be a bit shorter. Let me try quickly.

hodgestar: I've pushed a version that is based on `HyperlinkedIdentityField`. It's more magic but does use fewer lines of code.

hodgestar:
@KaitCrawford Ready for re-review.

KaitCrawford:
:+1:

PR #84, opened 2016-07-07
add download as csv with filter

codiebeulaine:


mentorship/mentor/admin.py
```diff
@@ -5,15 +5,88 @@
...
+            ('This Year', _('This Year')),
+        )
+
+    def queryset(self, request, queryset):
+        """
+        Returns the filtered queryset based on the value
+        provided in the query string and retrievable via
+        `self.value()`.
+        """
+        global DATE_RANGE
```
miltontony: Definitely -100 on this. Please find a way of doing this without `global`

codiebeulaine: @miltontony yeah, haven't implemented your suggestion. Will do so once it works

mentorship/mentor/admin.py
```diff
@@ -5,15 +5,88 @@
...
+    else:
+        next_month = today.replace(month=today.month + 1, day=1)
+    next_year = today.replace(year=today.year + 1, month=1, day=1)
+
+    if DATE_RANGE == 'Any Date' or DATE_RANGE == '':
+        return queryset
+    if DATE_RANGE == 'Today':
+        return queryset.filter(**{field_name: today})
+    if DATE_RANGE == 'Past 7 Days':
+        return queryset.filter(field_name__gte=(today - datetime.timedelta(
```
codiebeulaine: @hodgestar @miltontony the problem I'm having is here, I'm not sure how to reference the dynamic field name as well as the gte filter

mentorship/mentor/admin.py
```diff
@@ -5,15 +5,88 @@
 from django.db.models import Q
 from django.utils.html import format_html
 from django.core.urlresolvers import reverse
+import csv
```
hodgestar: Can we put all the standard library imports at the top?

hodgestar:
Just noting that we closed this because it turns out the actual feature request is to export calls and messages (see #85).

PR #85, opened 2016-07-07
add download as csv to calls and messages

codiebeulaine:


mentorship/mentor/admin.py
```diff
@@ -1,20 +1,37 @@
...
 from django.core.urlresolvers import reverse
+from django.http import HttpResponse
 
 from .models import (Mentor, Mentee, Call, Message, ScheduledCall, Task, Event,
                      Feedback, NotificationSettings, CustomPushNotification)
 
 
+def download_as_csv(mentorship_statstics_admin, request, queryset):
+    opts = queryset.model._meta
+    response = HttpResponse(content_type='text/csv')
```
hodgestar: Maybe we should add `; charset=utf-8` to the `content_type` to be clear that it's meant to be UTF-8 encoded.

mentorship/mentor/admin.py
```diff
@@ -1,20 +1,37 @@
...
+    response['Content-Disposition'] = 'attachment;filename=export.csv'
+    writer = csv.writer(response)
+    field_names = [field.name for field in opts.fields]
+    writer.writerow(field_names)
+    for obj in queryset:
+        writer.writerow([getattr(obj, field) for field in field_names])
```
hodgestar: We probably need to be a bit careful here and explicitly encode rows to UTF-8 ourselves. So something like:
```python
writer.writerow([
    unicode(getattr(obj, field)).encode('utf-8') for field in field_names
])
```

hodgestar:
Looking good so far. We'll probably have to make the export a bit more custom in order to handle the following fields properly:
* Calls:
  * Handler number (mentor)
  * Beneficiary number (mentee)
  * IVR ignored
  * IVR completed
  * IVR Results (Need to determine how these will be displayed)
* Messages:
  * Sender
  * Recipient

mentorship/mentor/admin.py
```diff
@@ -1,26 +1,75 @@
...
 from .models import (Mentor, Mentee, Call, Message, ScheduledCall, Task, Event,
                      Feedback, NotificationSettings, CustomPushNotification)
 
 
+def download_as_csv(model_admin, request, queryset):
+    opts = queryset.model._meta
+    response = HttpResponse(content_type='text/csv; charset=utf-8')
+    response['Content-Disposition'] = 'attachment;filename=export.csv'
+    writer = csv.writer(response)
+    field_names = [field.name for field in opts.fields]
```
hodgestar: It doesn't look like we're using `field_names` any more so we can remove it and `opts`.

mentorship/mentor/admin.py
```diff
@@ -1,26 +1,75 @@
...
+
+    def call_duration(obj):
+        if obj.end_time is not None:
+            return obj.end_time - obj.start_time
+        return "Ongoing"
+
+    def ivr_complete(obj):
```
hodgestar: I think we can delete these copies of `ivr_complete` and `call_duration` and just use `model_admin.ivr_complete` and `model_admin.call_duration`?

mentorship/mentor/admin.py
```diff
@@ -1,26 +1,72 @@
...
+
+    @classmethod
+    def call_duration(cls, obj):
         if obj.end_time is not None:
             return obj.end_time - obj.start_time
         return "Ongoing"
 
+    @classmethod
+    def ivr_complete(cls, obj):
+        return Feedback.objects.get(
```
hodgestar: We need to handle the case where there is currently no Feedback object associated with the call. In that case we can just return `None`, I think.

hodgestar:
Let one last comment but otherwise looks good.

We need to figure out how to handle the IVR results, but I think it's best to put that in a separate PR since we'll need to discuss it with people.

mentorship/mentor/tests/test_admin.py
```diff
@@ -147,6 +147,13 @@ def test_calls_download_as_csv(self):
         mentor1, mentee1 = utils.mk_mentor_mentee_pair('mentor1', 'mentee1')
         mentor2, mentee2 = utils.mk_mentor_mentee_pair('mentor2', 'mentee2')
         call = utils.mk_call(caller=mentor1, receiver=mentee1)
+        response = download_as_csv_calls(CallAdmin, None, Call.objects.all())
```
hodgestar: Could we make this a separate test rather than putting it into an existing test? Apologies for all the nitpicking. :/

hodgestar:
:+1: once Travis is happy! :D :cake:

PR #87, opened 2016-07-11
Add 'complete' field to mentorship feedback.

hodgestar:
We need to know whether feedback responses were partial or complete. Currently the feedback response Django model and the API don't capture this information explicitly.


hodgestar:
@codiebeulaine Ready for review.

I'm going to add the migrations in a separate PR because there are a few extra unrelated migrations needed too.

codiebeulaine:
👍

PR #90, opened 2016-07-11
Use Feedback model for IVR feedback statistics.

hodgestar:
Currently the statistics admin uses the Task model for IVR feedback statistics. This is incorrect and we should use the Feedback model's new `complete` field instead.


hodgestar:
@codiebeulaine Ready for review.

codiebeulaine:
👍 🍰

PR #92, opened 2016-07-25
Add id field for profile API.

hodgestar:
Currently the profile API does not include the user id, which makes updating profiles difficult. We should include the 'id' in API responses.


hodgestar:
@peterp Ready for review.

peterp:
👍

PR #97, opened 2016-08-12
add category API

codiebeulaine:


mentorship/mentor/urls.py
```diff
@@ -19,6 +19,11 @@
 urlpatterns = [
     # API urls
     url(r'^mentor/', include(router.urls)),
+    url(r'^category/', views.CategoryViewSet.as_view({'get': 'list'}),
+        name="category"),
+    url(r'^category/(?P<pk>\d+)/$',
+        views.CategoryViewSet.as_view({'get': 'list-detail'}),
+        name="category-detail"),
```
KaitCrawford: You can add a router in the section above this which will automatically create these (although the names might be different). `router.register(r'category', views.CategoryViewSet)` should work.

mentorship/mentor/views.py
```diff
@@ -74,6 +74,22 @@ def perform_update(self, serializer):
...
+                      viewsets.GenericViewSet):
+    """
+    This viewset automatically provides `list`, `create`, `detail` and `update`
+    actions.
+    """
+    serializer_class = CategorySerializer
+    permission_classes = (permissions.IsAuthenticated,)
+
+    def get_queryset(self):
+        return Category.objects.all()
```
miltontony: This should be `Category.objects.live()`

mentorship/mentor/views.py
```diff
@@ -74,6 +74,22 @@ def perform_update(self, serializer):
...
+
+    def get_object(self):
```
KaitCrawford: Using the router means you shouldn't need to specify these two methods here. It's done automatically.

mentorship/mentor/serializers.py
```diff
@@ -1,9 +1,15 @@
...
 from mentor.models import Call, Message, Event, Task, ScheduledCall, Mentee, \
-    Mentor, NotificationSettings
+    Mentor, NotificationSettings, Category
 from django.contrib.auth.models import User
 
 
+class CategorySerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Category
+        fields = ('title', 'image', 'color', 'about', 'goal')
```
KaitCrawford: Maybe add `'id'` to the fields so that the list view has the information necessary to drill down to the detail view (otherwise people using it won't have any way of getting to the detail view because they won't know the pk)

mentorship/mentor/tests/test_views.py
```diff
@@ -57,6 +59,40 @@ def mk_mentor(self, name='mentor', mentee=None):
...
+        category = Category(title='Test Category', slug='test-category')
+        self.main.add_child(instance=category)
+        category.save_revision().publish()
+        category2 = Category(title='Test Category2', slug='test-category2')
+        self.main.add_child(instance=category2)
+        category2.save_revision().publish()
+        self.assertEquals(Category.objects.live().child_of(
+            self.main).count(), 2)
+        self.assertEquals(Category.objects.all().count(), 2)
+        response = self.client.get(reverse('category'))
```
KaitCrawford: This url name will probably become `category-list` due to using the router to generate it

KaitCrawford:
@codiebeulaine I just added a suggestion but it looks good. Tests should be easy to fix :)

codiebeulaine:
@KaitCrawford ready for review

mentorship/mentor/tests/test_views.py
```diff
@@ -23,9 +23,10 @@
 from django.contrib.auth import views as auth_views
 from django.contrib.auth.models import User
 from django.contrib.auth.tokens import default_token_generator
+from mentor.tests.base import TestCaseMixin
 
 
-class MentorViewTests(TestCase):
+class MentorViewTests(TestCase, TestCaseMixin):
```
KaitCrawford: what is the `TestCaseMixin` for? (for my learning)

mentorship/mentor/urls.py
```diff
@@ -15,6 +15,7 @@
 router.register(r'schedule', views.ScheduledCallViewSet, "mentor-schedule")
 router.register(r'notification-settings', views.NotificationSettingsViewSet,
                 "mentor-notifications")
+router.register(r'category', views.CategoryViewSet, "category")
```
KaitCrawford: You don't need the third argument here. It's optional and is used to change the url name. (ie the schedule urls would have names like `schedule-list` and `schedule-detail` but because of the third argument they instead have names like `mentor-schedule-list`)

mentorship/mentor/tests/test_views.py
codiebeulaine: it is not needed here so just took it out, but it allows you to created a main page in wagtail really easily with self.mk_main()

KaitCrawford:
@codiebeulaine I've left a question and a nit-pick but :+1: when the tests pass

mentorship/mentor/urls.py
codiebeulaine: it fails without the name there

codiebeulaine:
@KaitCrawford it fails without that name there so left it in, still 👍 ?

KaitCrawford:
:confused: clearly I misunderstood something. :+1: cake will come later (maybe)

codiebeulaine:
woot!

PR #110, opened 2016-09-20
Feature/update notifications

codiebeulaine:


codiebeulaine:
@miltontony @justinvdm @KaitCrawford ready for review

mentorship/mentor/models.py
```diff
@@ -113,6 +113,9 @@ def ended(self, end_time=None, partner_id=None, recording_url=None,
                              event_type="end_call",
                              relevant_mentor=self.caller.mentor,
                              object_id=self.pk)
+        if self.caller.mentor.notification_settings.call_ended:
+            send_notification_to_gcm(self.caller.mentor, "Call ended",
+                                     self.end_time, self.pk)
```
justinvdm: I just realised now that we probably don't want a native notification for this type, so I think the `notification` field should be left out for this type. Maybe we should extend `send_notification_to_gcm` to include an optional flag defaulting to `True` for whether or not to use the notification field, and set it to `False` here, or something similar?

justinvdm: (Sorry for only noticing this now)

justinvdm:
Left a comment on a thing I realised, looks good :)

mentorship/mentor/models.py
codiebeulaine: cool. Now worries. updating now

justinvdm:
:+1:wtih

mentorship/mentor/models.py
```diff
@@ -231,7 +235,8 @@ def notify_callee(self, message):
     def remind_caller(self):
         time_string = self.call_time.strftime("%H:%M")
         text = "You have a call scheduled for " + time_string + "."
-        return send_notification_to_gcm(self.caller, "Call reminder", text)
+        return send_notification_to_gcm(
+            self.caller, "Call reminder", text, self.pk)
```
justinvdm: Two things I just noticed:
  - For the `type`, could we use more `programmatically_usable`, like the [types in the gist](https://gist.github.com/justinvdm/c2b6c05fb5f32bfdc3074f78412a128b#types)?
  - we are going to need separate types for the different reminders (e.g. CALL_STARTING_HOUR, CALL_STARTING_10_MIN, CALL_STARTING_1_MIN)

justinvdm: (if it would be more consistent with the event type naming, we could lowercase the types (e.g. `call_ended` rather than `CALL_ENDED`, happy either way)

PR #120, opened 2016-09-27
Feature/implement mentor profile filter page

codiebeulaine:
@miltontony ready for review cc @justinvdm


mentorship/mentor/serializers.py
```diff
@@ -287,19 +287,32 @@ class ProfileSerializer(serializers.ModelSerializer):
...
+            'date_joined', 'latest_post', 'latest_post_date')
 
     def get_num_posts(self, obj):
         return Post.objects.filter(user=self.context['request'].user).count()
 
+    def get_date_joined(self, obj):
+        return obj.user.date_joined.date()
+
+    def get_latest_post(self, obj):
+        if Post.objects.filter(
```
miltontony: Please change this to 
```python
post = Post.objects.filter(user=obj.user).last()
if post:
    return post.comment
```
This ensures that we're only making 1 call to the database and not 2.
The way you have it now, it will make a DB call on `.exists()` as well as on `.last()`

mentorship/mentor/views.py
```diff
@@ -565,10 +565,13 @@ class Meta:
...
                      viewsets.GenericViewSet):
-    queryset = Mentor.objects.all()
+    queryset = Mentor.objects.all().order_by('-latest_post_date')
     serializer_class = ProfileSerializer
     permission_classes = (permissions.IsAuthenticated,)
-    filter_class = ProfileFilter
+    filter_backends = (filters.OrderingFilter, filters.DjangoFilterBackend,)
+    filter_class = ProfileEmailFilter
+    ordering_fields = (
+        'user__username', 'user__date_joined')
```
miltontony: Please confirm this - I think the default order should be alphabetic by `Firstname` and not `username`

miltontony:
2 small changes, otherwise this is good.

mentorship/mentor/serializers.py
```diff
@@ -308,10 +308,9 @@ def get_date_joined(self, obj):
         return obj.user.date_joined.date()
 
     def get_latest_post(self, obj):
-        if Post.objects.filter(
-                user=obj.user).exists():
-            return Post.objects.filter(
-                user=obj.user).last().comment
+        post = Post.objects.filter(user=obj.user).last()
```
miltontony: Can you please enforce the ordering here - so that you can be sure that it's the latest post. Add `.order_by('-submit_date').first()`

codiebeulaine: done

mentorship/mentor/views.py
codiebeulaine: done

mentorship/mentor/serializers.py
codiebeulaine: done

miltontony:
👍 🍰 once travis

PR #122, opened 2016-09-28
return 200 even if no update and add activity to scheduled call update

codiebeulaine:
@justinvdm ready for review


mentorship/mentor/tests/test_views.py
```diff
@@ -1434,6 +1412,30 @@ def test_schedule_detail_put(self):
...
+        activity.save_revision().publish()
+        data = json.dumps({
+            'activity': activity.pk,
+            'call_time': new_call_time.strftime('%Y-%m-%dT%H:%M')})
+        response = self.client.put(url, data, 'application/json')
+        self.assertEqual(response.status_code, 200)
+        self.assertTrue(ScheduledCall.objects.filter(
+            activity=activity.pk).exists())
+
+        # test that response is 2 even if no update
```
justinvdm: Should this say `200`?

mentorship/mentor/views.py
```diff
@@ -311,23 +310,32 @@ def perform_update(self, serializer):
...
+        if 'activity' in data:
+            queryset = ScheduledCall.objects.filter(
+                caller=caller, call_time=data['call_time'],
+                activity=data['activity']).exists()
+        else:
+            queryset = ScheduledCall.objects.filter(
+                caller=caller, call_time=data['call_time']).exists()
+        if queryset:
+            return HttpResponse(
+                'There is already a call for this date or with this actvity.')
```
justinvdm: I probably didn't explain this clearly. I'd like us to respond with `200` for this case also. To the client's point of view, their update request should succeed regardless of whether anything or nothing changed as a result of their request.

codiebeulaine: @justinvdm it does return a 200 response

codiebeulaine: @justinvdm made some changes now

mentorship/mentor/tests/test_views.py
codiebeulaine: yep

mentorship/mentor/views.py
justinvdm: Ah, sorry, misread :) This still needs changing though: it should respond with the model's data (the same as a normal update). Thought that was implied by the `200` but I should've made that clearer, sorry.

codiebeulaine: @justinvdm I took that out now, look good?

justinvdm: yep :) thanks

justinvdm:
👍 wtih

PR #123, opened 2016-09-28
calling fcm instead of gcm

codiebeulaine:


codiebeulaine:
@miltontony ready for review

miltontony:
👍 🍰 once travis!

mentorship/mentor/utils.py
```diff
@@ -34,25 +34,25 @@ def send_reset_email(to_email, context):
...
-def send_notification_to_gcm(
+def send_notification_to_fcm(
         mentor, title, content, object_id, use_notification=True):
-    try:
-        gcm = GCM(settings.GCM_API_KEY)
-    except AttributeError:
-        return False
 
+    if not hasattr(settings, 'FCM_DJANGO_SETTINGS'):
+        return False
```
justinvdm: We don't seem to be using this setting. Should we be accessing the auth key off of it?

codiebeulaine: The Device checks for it in the settings and uses it behind the scenes

PR #128, opened 2016-09-30
send separate notifications

codiebeulaine:
@justinvdm ready for review


mentorship/mentor/tests/test_models.py
```diff
@@ -327,32 +327,30 @@ def test_send_notification(self):
...
+                    mentor.pk, use_notification=False)
                 mock_method.assert_called_once_with(
                     data={
                         'type': 'Test notification',
                         'payload': {
                             'content': 'This is a test',
-                            'object_id': 2}},
-                    body='This is a test',
-                    title='Test notification')
+                            'object_id': 2}})
```
justinvdm: Do we want to add a test for when `use_notification` is `True`, and check that we send out two notifications for that case? We could probably do that with something like:

```py
# at the top
from mock import call

# in a test
mock_method.assert_has_calls(
  call(...),  # custom data send
  call(...))  # notification data send
```

justinvdm:
Looks good :) just the one comment

justinvdm:
:+1:wtih 🍰

PR #138, opened 2016-10-17
Feature/add activity to call serializer

codiebeulaine:
@justinvdm ready for review


justinvdm:
:+1:wtih

I think [this](https://github.com/praekelt/mentorship/blob/3db20586c3040384a6971152cd6de33d2d64fb5e/mentorship/mentor/serializers.py#L110-L111) needs to return `activity.id`, but that is a separate issue, so I'm guessing that is planned for a different PR?

codiebeulaine:
@justinvdm yeah :)

PR #142, opened 2016-10-19
Support adding a scheduled call to a call

justinvdm:
The app currently doesn't have way of finding out which scheduled call is associated with a call. We sometimes have the context to figure this out, but sometimes we don't. Regardless, the simpler and less work-around-y solution is for the api to support this.


justinvdm:

codiebeulaine:
@justinvdm 👍

PR #143, opened 2016-10-19
localizing time in remind caller message

codiebeulaine:


justinvdm:
Looks good, but I think we need a test. You mentioned that is difficult because of the way we are mocking things, but I'm not really following. I was thinking we could do something like localizing the time [in this test](https://github.com/praekelt/mentorship/blob/develop/mentorship/mentor/tests/test_models.py#L132)  (and similar tests, if there are) to be in IST. I think I'm missing something though, could you elaborate on the mocking issue?

justinvdm:
:+1:wtih 🍰

PR #144, opened 2016-10-25
call end_call method when starting a call

codiebeulaine:


codiebeulaine:
@justinvdm @KaitCrawford ready for review

mentorship/comms/views.py
```diff
@@ -42,12 +42,13 @@ def post(self, request, format=None):
...
-            logging.error(
-                "end_time can't be parsed into a datetime", exc_info=True)
-            raise ValidationError("end_time can't be parsed into a datetime")
+        # try:
+        #     end_time = datetime.strptime(
+        #       end_time[:-6], "%Y-%m-%dT%H:%M:%S.%f")
+        # except KeyError:
+        #     logging.error(
+        #         "end_time can't be parsed into a datetime", exc_info=True)
+        #     raise ValidationError("end_time can't be parsed into a datetime")
```
justinvdm: Why are we commenting this out?

codiebeulaine: this was added in case timezone info was added to the datetime, but it won't be added since we are using strftime

mentorship/comms/models.py
```diff
@@ -32,17 +33,21 @@ def send_message(self, sender_id, recipient_id, content):
...
+
     def make_call(self, caller_id, receiver_id):
+        from mentor.models import Call
+
         # the url used to connect to the api
         url = self.call_url
-
+        call = Call.objects.all().last()
         # the post data the url expects to receive
+        end_time = (datetime.now().strftime('%Y-%m-%dT%H:%M:%S') + '.948000')
```
justinvdm: What is with the `.948000`? Are we just choosing some arbitrary time in the future?

codiebeulaine: @justinvdm yip

mentorship/comms/models.py
```diff
@@ -32,17 +33,21 @@ def send_message(self, sender_id, recipient_id, content):
...
 # Example API. This does not actually connect to anything!!
 class ExampleApi(Api):
+
     def make_call(self, caller_id, receiver_id):
+        from mentor.models import Call
+
         # the url used to connect to the api
         url = self.call_url
-
+        call = Call.objects.all().last()
```
justinvdm: Wondering if it would ever be the case that the last call wasn't actually the relevant call here. I guess since this stuff isn't async, and since we are calling `make_call()` immediately after creating this isn't an issue though.

mentorship/comms/views.py
KaitCrawford: Please remove this rather than commenting it out. Thanks

justinvdm: Maybe it will be added by onion dev though, so I'd rather we didn't remove this. If this code isn't a no-op for the dummy api, could we maybe just get the dummy api to send something in the format we are expecting?

mentorship/comms/models.py
KaitCrawford: You can probably remove that since you've removed the time parsing in the view

mentorship/comms/views.py
codiebeulaine: @justinvdm maybe @KaitCrawford can explain better. We do send it in the correct format

mentorship/comms/models.py
codiebeulaine: @justinvdm my understanding is that this is just for testing. so we just took the last call created

mentorship/comms/views.py
KaitCrawford: @justinvdm this endpoint is separate from the OnionDev endpoint (don't get me started on why). It's intended to be the default endpoint but isn't currently used by anything.

justinvdm: Ah, thanks. Hmm, I'm wondering if this is a good idea to be using this endpoint then. Seems like it would make for more thorough tests for us to be using the endpoint actually to be used in production, rather than using dead code paths we should maybe be removing (unless I'm misunderstanding and it will eventually be used by something).

To not hold us up, I'm alright with us continuing with this approach and just removing this instead of commenting it out, but it doesn't feel like first prize.

justinvdm:
If it is at all possible and it won't take that much longer to do, I'd rather we used the actual endpoint onion dev uses. If this isn't the case, then I'm alright with this landing once travis is happy, the commented out code has been removed, and the corresponding `.948000` thing has been removed.

mentorship/comms/models.py
```diff
@@ -32,17 +33,29 @@ def send_message(self, sender_id, recipient_id, content):
...
     def make_call(self, caller_id, receiver_id):
         # the url used to connect to the api
         url = self.call_url
+        handler = caller_id
+        end_time = datetime.now().strftime('%Y-%m-%d %H:%M:%S.%f')
+
+        conversationEndTime = end_time
+        conversationStartTime = end_time
+        partner_id = 1
+        recording_url = 'http://localhost:8000/'
```
justinvdm: Is this a field that we expect from the oniondev api that we don't actually use, so we are stubbing it out here?

justinvdm:
Left a few minor comments, looks good :) 👍  when those are resolved and travis is happy

mentorship/comms/models.py
codiebeulaine: @justinvdm yip

PR #148, opened 2016-11-03
changed image field with failing tests

codiebeulaine:


codiebeulaine:
:+1:

mentorship/mentor/views.py
```diff
@@ -623,14 +626,24 @@ def image_put(self, request, *args, **kw):
...
+            mentor.profile_pic.delete()
+
+        image_file = ImageFile(profile_pic_data)
+        image = Image.objects.create(
+            title=request.user.get_full_name() or request.user.username,
+            file=image_file)
+        image.save()
+        mentor.profile_pic = image
+        mentor.save()
+        print mentor.profile_pic, '()', image, profile_pic_data
```
justinvdm: Stray print?

codiebeulaine: taking out

mentorship/mentor/models.py
```diff
@@ -57,8 +56,13 @@ def profile_pic_path(instance, filename):
...
                                   blank=True)
-    profile_pic = models.ImageField(
-        upload_to=profile_pic_path, null=True, blank=True)
+    profile_pic = models.ForeignKey(
+        'wagtailimages.Image',
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name='+'
+    )
```
justinvdm: If we are using the same field, do we know how migrations for this will work? Just wondering if this means we need to clear away all existing profiles.

mentorship/mentor/serializers.py
```diff
@@ -328,6 +316,9 @@ def get_latest_post(self, obj):
         if post:
             return post.comment
 
+    def get_profile_pic(self, obj):
+        return get_image_url(obj.profile_pic)
```
justinvdm: Does this mean the profile pic endpoint is public? Maybe not too big deal if other mentors can see other mentors profile pics, but it does mean people without logins can potentially get people's profile pics if this is the case.

mentorship/mentor/tests/test_views.py
```diff
@@ -1872,47 +1875,49 @@ def test_profile_detail_put(self):
...
+        self.client.put(url, **self.mk_multipart(data))
 
-    def mk_profile_pic_detail_url(self, mentor_pk):
-        return "%s%s" % (
-            reverse(
-                self.url_name_base + 'profile-detail',
-                kwargs={'pk': mentor_pk}),
-            'image/')
+        updated_mentor = Mentor.objects.get(pk=mentor.pk)
+        print updated_mentor.profile_pic, '<---', mentor.user.get_full_name()
```
justinvdm: stray print?

mentorship/mentor/serializers.py
codiebeulaine: @justinvdm this works the same way as the other images we use in wagtail like the category and activity images

PR #149, opened 2016-11-04
Use wagtail image field for profile pic - fixes

codiebeulaine:


justinvdm:
:+1:

mentorship/mentor/migrations/0029_image_field.py
```diff
@@ -0,0 +1,20 @@
...
+        ('mentor', '0028_call_scheduled_call'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='mentor',
+            name='profile_pic',
+            field=models.ForeignKey(related_name='+', on_delete=django.db.models.deletion.SET_NULL, blank=True, to='wagtailimages.Image', null=True),
+        ),
+    ]
```
justinvdm: Not sure if this is something that will automagically work for existing rows. We should check if existing profiles are still queryable and updatable in qa before deploying to prod.

PR #151, opened 2016-11-04
Stringify datetime for call end notification when sending to fcm

justinvdm:
At the moment we give the fcm client library we are using a `datetime()`, which it tries to JSON serialize.


codiebeulaine:
👍

PR #155, opened 2016-12-12
adjusting message model for chat

codiebeulaine:
@justinvdm ready for review


mentorship/mentor/serializers.py
```diff
@@ -173,11 +173,19 @@ class Meta:
...
         model = Message
-        fields = ('id', 'time_sent', 'sender', 'recipient', 'content')
-        read_only_fields = ('sender', 'recipient')
+        fields = (
+            'id', 'time_sent', 'sender', 'recipient', 'content', 'direction',
+            'message_type')
+        read_only_fields = ('sender', 'recipient', 'direction', 'message_type')
+
+    def get_direction(self, obj):
+        if self.context['request'].user.pk is obj.sender.user.pk:
```
justinvdm: I think it is better practise and less error-prone to use `==` for int comparisons, can we use `==` here?

codiebeulaine: sure

justinvdm:
🍰 👍 wtih!

Wondering if we need to split out the user and system out messages the stats section of the admin UI since it is still showing all sent messages (user and system), when people might just be expecting user. Can be another PR though.

codiebeulaine:
@justinvdm i do have a ticket for admin clean up, will add that to the list

PR #159, opened 2016-12-14
Block on OnionDev api calls

justinvdm:
At the moment we perform these calls async via celery. We need to block on these to avoid creating messages and telling the client side we've completed sending for cases where failures have happened (similarly for making calls).


justinvdm:

codiebeulaine:
@justinvdm 👍

PR #162, opened 2016-12-15
allow custom filter_spec in url

miltontony:


miltontony:
ready for review

justinvdm:
:+1:wtih but would be cool if we could test that we are supporting the `filter_spec` param

justinvdm:
Ah, was expecting that we give `get_rendition()` the correct filter spec params based on the query param given, but maybe that is overkill. :+1:wtih

PR #164, opened 2016-12-19
More logging

justinvdm:


mentorship/mentorship/settings.py
```diff
@@ -190,6 +200,8 @@
         }
     },
     'loggers': {
+        'comms': APP_LOGGER,
+        'mentor': APP_LOGGER,
         'django': {
             'level': 'DEBUG',
             'handlers': ['console', 'sentry'],
```
justinvdm: @miltontony you probably have a much better idea how django/python logging works. Could you take a look at these settings to check if I'm doing anything that shouldn't be done?

justinvdm:
@codiebeulaine @miltontony I tried to make sure we aren't logging credentials, but I have assumed that it is okay if user-related content is included in our logs (msisdns, and sms content). Scrubbing the logs of these is going to be a bit more involved. Is this okay for this project?

mentorship/mentorship/settings.py
miltontony: This looks good - the only way to know for sure is to test it locally and see if it's logging.. sometimes it takes a few times of fiddling to get the logging to work 💯 %

PR #165, opened 2016-12-20
use id to end calls

codiebeulaine:


mentorship/mentor/models.py
```diff
@@ -99,10 +99,16 @@ class Call(models.Model):
...
+    call_id = models.CharField(
+        max_length=255, blank=True, null=True, default='')
 
     def dial(self):
         api = comms_models.Api.objects.last()
-        api.make_call(self.caller.contact_number, self.receiver.contact_number)
+        response = api.make_call(
+            self.caller.contact_number, self.receiver.contact_number)
+        if response:
+            self.call_id = response.get('scheduledCallId')
```
justinvdm: iirc, they use `scheduled_call_id` on this side (and `scheduledCallId` when ending calls). Sorry, realised I didn't tell you about that.

mentorship/comms/views.py
```diff
@@ -109,15 +109,25 @@ def post(self, request, format=None):
...
             log.error(
                 'This call has already ended.', extra={'stack': True})
             raise ValidationError('This call has already ended.')
         end_time = datetime.now()
 
-        call.ended(end_time, partner_id, recording_url, partner_start_time,
-                   partner_end_time)
+        last_call.ended(
+            end_time, partner_id, recording_url, partner_start_time,
+            partner_end_time)
```
justinvdm: Nitpicking a bit, but I think `last_call` is a bit of a misleading name here, since it is more likely that the call is the one associated to the id, rather than the last call (last call is just a fallback). Maybe we should rather change the code so that we are still using `call` as the name for the relevant call (regardless of whether it is the last call, or the call with the associated id).

mentorship/comms/views.py
```diff
@@ -109,15 +109,25 @@ def post(self, request, format=None):
...
-        if call.end_time is not None:
+        last_call = mentor_models.Call.objects.filter(caller=caller.pk).last()
+        try:
+            call_id = data['scheduledCallId']
+            call = mentor_models.Call.objects.filter(call_id=call_id).first()
+            if call:
+                last_call = call
+        except ValueError:
+            logging.error(
+                "Invalid call id",
```
justinvdm: Invalid call id implies the response did in fact include a call id. I think it would be clearer if we mentioned here that the response did not contain a scheduled call id. It would also be useful if the message mentioned that we are falling back to the last call that does not have an end time.

mentorship/comms/views.py
codiebeulaine: @miltontony hahaha @justinvdm changing now

mentorship/comms/views.py
codiebeulaine: 👍

mentorship/comms/views.py
```diff
@@ -109,15 +109,25 @@ def post(self, request, format=None):
             raise ValidationError(
                 "'conversationEndTime' is in the incorrect format."
             )
-        call = mentor_models.Call.objects.filter(caller=caller.pk).last()
-        if call.end_time is not None:
+        last_call = mentor_models.Call.objects.filter(caller=caller.pk).last()
+        try:
+            call_id = data['scheduledCallId']
+            call = mentor_models.Call.objects.filter(call_id=call_id).first()
```
justinvdm: Wondering if we should rather use `.get()` instead of `.filter(...).first()`, and log an error if it throws an exception. If we get the same id twice, it means there is a bug on onion dev's side, so we probably want to know about it.

I'm not sure if there is a way we can recover from that case gracefully, if we use the first matching call anyways, it may put us in a weird erroneous state we haven't thought about, so my vote is we log an error don't end anything.

mentorship/comms/views.py
```diff
@@ -109,15 +109,27 @@ def post(self, request, format=None):
...
+            if call:
+                call_to_end = call
+        except ValueError:
+            logging.error(
+                "Response does not contain scheduledCallId. "
+                "The last call made will be ended.",
+                exc_info=True)
```
justinvdm: :+1: to message :) just realised we should be using `log` instead of logging now that #164 has landed.

mentorship/comms/views.py
codiebeulaine: @justinvdm 👍

mentorship/comms/views.py
```diff
@@ -109,15 +109,31 @@ def post(self, request, format=None):
...
                 'This call has already ended.', extra={'stack': True})
             raise ValidationError('This call has already ended.')
         end_time = datetime.now()
 
-        call.ended(end_time, partner_id, recording_url, partner_start_time,
-                   partner_end_time)
+        call_to_end.ended(
+            end_time, partner_id, recording_url, partner_start_time,
+            partner_end_time)
         return HttpResponse(status=200)
```
justinvdm: 💯  to the naming, error handling, logging, and everything here :)

mentorship/comms/tests/test_comms_models.py
```diff
@@ -125,6 +135,28 @@ def test_onion_dev_make_call(self):
...
+                    u'feedbackSubscriptionStatus': True,
+                    u'conversationId': u'5596',
+                    u'conversationStartTime': u'2016-12-15 19:32:12.528267',
+                    u'handler': u'5678',
+                    u'cdr': {
+                        u'start_time': u'2016-12-15 19:31:43.571941',
+                        u'id': u'28668022', u'callerid': u'918800098471'},
+                    u'recordingUrl': u'http://voice2.gramvaani.org/fsmedia/re \
+                        cordings/889/1081938.mp3',
+                    u'scheduled_call_id': u'2145100'}):
```
justinvdm: On this side, we only get the scheduled call id, so we can probably remove the other fields.

mentorship/comms/views.py
```diff
@@ -109,15 +109,31 @@ def post(self, request, format=None):
...
+                "The last call made will be ended.",
+                exc_info=True)
+        try:
+            call_to_end = mentor_models.Call.objects.get(call_id=call_id)
+        except ObjectDoesNotExist:
+            log.error(
+                "Call does not exist. The last call made will be ended.",
+                exc_info=True)
+            call_to_end = mentor_models.Call.objects.filter(
+                caller=caller.pk).last()
```
justinvdm: Sorry, only noticed another case here now. This handles when we don't have a call with the give id. We also need to handle the case where there are multiple matches for a call id. There, I think the best we can do is log and error and not end any call.

justinvdm: On second thought, for that cases, we could also probably fall back to the last unended call.

mentorship/comms/tests/test_comms_views.py
```diff
@@ -160,17 +160,24 @@ def test_call_end_handler_onion_dev(self):
...
+            caller=call.caller, receiver=call.receiver)
         self.assertIsNone(call.end_time)
+        self.assertIsNone(new_call.end_time)
         response = self.client.post(url, data)
         self.assertEqual(response.status_code, 200)
         call = mentor_models.Call.objects.get(pk=call.pk)
+        self.assertIsNone(new_call.end_time)
         self.assertIsNotNone(call.end_time)
         self.assertNotEqual(call.end_time, end_time.astimezone(pytz.utc))
         self.assertEqual(call.partner_start_time,
```
justinvdm: Could we add tests for these cases:
  - No `scheduledCallId` in request
  - No call with the given `scheduledCallId` exists
  - Multiple calls with the given `scheduledCallId` exist

codiebeulaine: 👍

justinvdm:
Left a few more minor comments. Looks great :)

justinvdm:
@codiebeulaine could you take a look at the changes I made? If you're happy, 👍 wtih from me, feel free to merge if that is the case.

codiebeulaine:
👍

PR #167, opened 2016-12-20
return 200 response if mentee already created

codiebeulaine:
@justinvdm @miltontony ready for review


mentorship/mentor/views.py
```diff
@@ -596,6 +597,30 @@ class MenteeViewSet(mixins.ListModelMixin, mixins.RetrieveModelMixin,
...
+    def create(self, request, *args, **kwargs):
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        mentee_created = self.perform_create(serializer)
+        headers = self.get_success_headers(serializer.data)
+        if mentee_created:
+            return Response(
+                serializer.data, status=status.HTTP_201_CREATED,
+                headers=headers)
+        return Response(serializer.data, status=200, headers=headers)
```
justinvdm: I'm not too clued up on django rest framework, but I'm not sure if `serializer.data` will include information not given by the client side (e.g. the model `id`), since a `serializer.save()` didn't happen. Doing a `serializer.save()` would probably end up creating a new model each time for `create`'s case though, so I think we need to do something different here.

One thing that could work would be to manually instantiate the serializer with a model we return from `perform_create()`, and then use its `.data` for these responses.

mentorship/mentor/tests/test_views.py
```diff
@@ -1774,6 +1774,43 @@ def test_mentee_list(self):
...
+            "first_name": "mentee",
+            "last_name": "surname",
+            "contact_number": "1234"
+        }
+        response = self.client.post(
+            reverse('mentee-list'), data)
+        self.assertEqual(response.status_code, 201)
+        response = self.client.post(
+            reverse('mentee-list'), data)
+        self.assertEqual(response.status_code, 200)
```
justinvdm: Can we check in this test and the test above whether we return the model `id`? Would allow us to make sure we aren't just sending back the data the client sent to us.

justinvdm:
👍  if we're sure the client side doesn't need `id`s or other mentor data that they didn't send themselves.

PR #168, opened 2016-12-21
fixing call id tests

codiebeulaine:
@KaitCrawford ready for review


mentorship/comms/tests/test_comms_models.py
```diff
@@ -135,11 +135,16 @@ def test_onion_dev_make_call(self):
             mock_task.assert_called_with(api.call_url, expected_data,
                                          expected_header)
 
+    class MockResponse():
```
justinvdm: We should subclass `object` here (`MockResponse(object):`): http://stackoverflow.com/questions/54867/what-is-the-difference-between-old-style-and-new-style-classes-in-python

justinvdm:
Two minor comments, looks good :)

justinvdm:
:+1: from me when travis is happy

PR #169, opened 2016-12-21
only show messages sent with message type USER

codiebeulaine:
@justinvdm ready for review


justinvdm:
👍 wtih, but are we sure we don't want to display system message stats at all?

codiebeulaine:
@justinvdm we do, when you go into the messages outside of the stats, if we go to messages through the stats, the admin will click on total messages sent to mentor, so through that link I will only show the ones by user :)

PR #173, opened 2016-12-21
link to call in call note admin

codiebeulaine:
@justinvdm ready for review. Wha I'm doing here is allowing the admin to click on the call and take them to the call admin for that call


mentorship/mentor/admin.py
```diff
@@ -88,11 +88,27 @@ class FeedbackAdmin(admin.ModelAdmin):
...
+        return format_html(
+            '<a href="{url}">{value}</a>', url=url, value=value)
+
     list_display = (
-        'mentor', 'call', 'reflection', 'mentee_state',
+        'mentor', 'Call', 'reflection', 'mentee_state',
         'objective_achieved', 'activity_helpful', 'call_quality')
     list_filter = ['mentor', 'objective_achieved']
 
+    def Call(self, obj):
```
miltontony: functions should be lowercase. If you want to show a different name, use `short_description`

codiebeulaine: 👍

mentorship/mentor/admin.py
```diff
@@ -88,11 +88,27 @@ class FeedbackAdmin(admin.ModelAdmin):
...
+        return url
+
+    def _link(self, value, route, **params):
+        url = self._link_url(route, **params)
+        return format_html(
+            '<a href="{url}">{value}</a>', url=url, value=value)
+
     list_display = (
-        'mentor', 'call', 'reflection', 'mentee_state',
+        'mentor', 'Call', 'reflection', 'mentee_state',
```
justinvdm: Nitpicking, but would it maybe be clearer to say `Show call`? It might not be as clear that clicking `Call` takes the user to the call. Not very important though.

Regardless, we should probably use [`short_description`](https://docs.djangoproject.com/en/1.10/ref/contrib/admin/#django.contrib.admin.ModelAdmin.list_display) instead of using a capital here.

justinvdm:
Left one minor comment. 👍  from me when that is resolved if travis is happy and we've tried this out locally.

mentorship/mentor/tests/test_admin.py
```diff
@@ -51,12 +51,19 @@ def test_call_notes_admin(self):
...
         self.assertContains(response, 'Mentee state')
         self.assertContains(response, 'Activity helpful')
         self.assertContains(response, 'Call quality')
 
+        response = self.client.get(reverse(
+            'admin:mentor_callnote_changelist'))
+        self.assertContains(
+            response,
+            '<td class="field-Call"><a href="/django-admin/mentor/call'
+            '/?id=1">Call</a></td>')
```
miltontony: Travis will fail..

miltontony:
👍 wtis

PR #174, opened 2016-12-21
Normalise contact numbers to include country codes

justinvdm:
We seem to get numbers without country codes from the matching side. Onion dev gives us things back with country codes added to the country codes, so we need to have our contact number fields normalised to include country codes for us to look up contacts by what we get back from Onion Dev.


codiebeulaine:
👍

PR #175, opened 2016-12-22
showing images in wagtail admin

codiebeulaine:
@justinvdm ready for review


mentorship/mentorship/urls.py
```diff
@@ -13,20 +13,46 @@
...
+    from django.contrib.staticfiles.urls import staticfiles_urlpatterns
+
+    urlpatterns += staticfiles_urlpatterns()
+    urlpatterns += static(
+        settings.MEDIA_URL + 'images/', document_root=os.path.join(
+            settings.MEDIA_ROOT, 'images'))
+    urlpatterns += patterns(
+        '',
+        (r'^favicon\.ico$', RedirectView.as_view(
+            url=settings.STATIC_URL + 'myapp/images/favicon.ico'))
```
justinvdm: Should `myapp` be here?

codiebeulaine: sorry forgot to take that out

mentorship/mentorship/settings.py
```diff
@@ -234,6 +234,11 @@
 
 WAGTAIL_SITE_NAME = ''
 
+WAGTAILIMAGES_FEATURE_DETECTION_ENABLED = False
+
+MEDIA_ROOT = join(PROJECT_ROOT, 'media')
+MEDIA_URL = '/media/'
```
justinvdm: Will this not cause the current pictures store at `/images` in qa and prd to no longer be found (unless we manually move them)? Wondering if we should rather just use the project route as the media route as we do at the moment. Thoughts?

codiebeulaine:
screen shot 2016-12-22 at 12 26 22 pm

justinvdm:
Two minor comments, looks good :)

mentorship/mentorship/settings.py
```diff
@@ -236,8 +236,8 @@
 
 WAGTAILIMAGES_FEATURE_DETECTION_ENABLED = False
 
-MEDIA_ROOT = join(PROJECT_ROOT, 'media')
-MEDIA_URL = '/media/'
+MEDIA_ROOT = join(PROJECT_ROOT)
```
justinvdm: Could we remove `join()`? I don't think we need it anymore.

justinvdm:
:+1: when travis is happy and that last minor comment of mine is resolved

PR #176, opened 2016-12-22
updating copy sent to mentee when call scheduled

codiebeulaine:
@justinvdm ready for review


justinvdm:
:+1:

codiebeulaine:
@justinvdm ready for review again

justinvdm:
First prize would be to change the tests to use hardcoded date values for the input dates and output messages, since that would make for much stronger tests, but if that is more work than its worth, happy with this landing.

justinvdm:
Would also be awesome if we could wrap up this timezone stuff into a util, but same case as above, happy with this landing if that is more work than it is worth.

justinvdm:
👍 wtih :cake::cake: util and tests look awesome :)

PR #179, opened 2017-01-03
Improve mentee sms copy

justinvdm:
Client asked for better copy


justinvdm:
@miltontony ready for review

mentorship/mentor/models.py
```diff
@@ -364,10 +368,26 @@ def notify_callee(self, message):
...
+        self.notify_callee(
+            "Your next mentoring call has been scheduled for %s" %
+            display_datetime(self.call_time))
+
+    def notify_callee_rescheduled(self, new_time):
+        self.notify_callee(
+            "Your mentoring call on %s has been moved to %s" %
+            (display_datetime(self.call_time), display_datetime(new_time)))
+
+    def notify_callee_canceled(self):
```
miltontony: typo `canceled`

mentorship/mentor/utils.py
```diff
@@ -80,5 +80,19 @@ def normalise_contact_number(num, country_code):
         return num
 
 
-def get_timezone_correct_time(time, timezone):
-    return time.astimezone(pytz.timezone(timezone))
+def get_timezone_correct_time(time, tz=None):
+    if tz is None:
+        tz = 'Asia/Kolkata'
```
miltontony: We should probably use the value from `settings` i.e `django.conf.settings.TIME_ZONE`

miltontony:
Tests failing because of _flake8_

mentorship/mentor/utils.py
justinvdm: Doesn't changing the `TIME_ZONE` change the timezone to use when reading or writing to/from the db? If so, I think it makes sense for that to stay`UTC`, and only use this timezone when we are wanting to display times in messages.

Happy to define a different setting for this though, if the aim is to keep these things in settings.

justinvdm:
@miltontony ready for re-review

mentorship/mentor/utils.py
miltontony: 🤔 Ok, let's use a custom settings value then. I need to do some investigation on how timezones glue together in Django - we're currently defaulting to UTC and the server time is probably SAST and we're using Asia/Kolkata for displaying.. I always worry about the final admin users using the site and what they actually see/use.

I'm happy to use a custom settings value for now.

miltontony:
👍 once timezone value moved to settings.

mentorship/mentor/utils.py
justinvdm: @miltontony did some reading :)

re db and server time, I think the server's timezone being SAST won't affect things in our case. [Postgres section in the timezone docs](https://docs.djangoproject.com/en/1.10/topics/i18n/timezones/#postgresql):
> The PostgreSQL backend stores datetimes as timestamp with time zone. In practice, this means it converts datetimes from the connection’s time zone to UTC on storage, and from UTC to the connection’s time zone on retrieval.
> As a consequence, if you’re using PostgreSQL, you can switch between USE_TZ = False and USE_TZ = True freely. The database connection’s time zone will be set to TIME_ZONE or UTC respectively, so that Django obtains correct datetimes in all cases. You don’t need to perform any data conversions.

To add to that, it seems that `TIME_ZONE` is only something you should use if you are using a db that doesn't support timezones, and postgres supports timezones. If you have `TIME_ZONE` set to `UTC` and `USE_TZ` set to `True` (what we have in settings), it seems like a no-op. Here's a snippet of the [docs for `TIME_ZONE`](https://docs.djangoproject.com/en/1.10/ref/settings/#time-zone):

> When USE_TZ is True and the database supports time zones (e.g. PostgreSQL), it is an error to set this option.
> When USE_TZ is False, it is an error to set this option.

re the times shown in the admin UI, in order for us to have those times shown in the user's timezone, it seems we need to set the [*current* time zone](https://docs.djangoproject.com/en/1.10/topics/i18n/timezones/#default-time-zone-and-current-time-zone) (not to be confused with the *default* timezone, which is based on `TIME_ZONE`) via [`activate()`](https://docs.djangoproject.com/en/1.10/ref/utils/#django.utils.timezone.activate).

**tldr** What we have at the moment works, except admin UI needs us to use [`activate()`](https://docs.djangoproject.com/en/1.10/ref/utils/#django.utils.timezone.activate()) if we want times displayed in user's timezone there.

PR #180, opened 2017-02-07
Support view type for looking up api resources via query params

justinvdm:
We need this so that the matching system can make requests to keep the two systems in sync without needing to store ids.

While storing and later re-using ids given back from this api on the matching side is an option, it has some shortcomings:
  - it implies a bit more logic on the matching side: the more of this we can avoid the better
  - emails are unique identifiers for users in the two systems, so needing to store another unique identifier is a little redundant
  - we'd need to first manually update existing users on the matching side to include these ids


justinvdm:

codiebeulaine:
👍

PR #184, opened 2017-02-08
Support querying mentors and mentees by matching platform id

justinvdm:
We need to support storing and querying by the ids for mentors and mentees on the matching side of the system to keep the two sides in sync (the current plan is to use email addresses, but this isn't robust enough, since emails are technically optional on the matching side).


justinvdm:
@codiebeulaine ready for review, though it would be good to discuss this change

codiebeulaine:
👍

mentorship/mentor/views.py
```diff
@@ -596,26 +597,23 @@ class MenteeViewSet(mixins.ListModelMixin, mixins.RetrieveModelMixin,
...
-    def perform_create(self, serializer):
-        user, user_created = User.objects.get_or_create(
-            first_name=serializer.validated_data.get('user')['first_name'],
-            username=serializer.validated_data.get('user')['username'],
-            last_name=serializer.validated_data.get('user')['last_name'],
-            email=serializer.validated_data.get('user')['email'])
+        if external_id is not None:
+            instance = Mentee.objects.filter(external_id=external_id).first()
+        else:
+            instance = Mentee.objects.filter(user__email=email).first()
```
justinvdm: @codiebeulaine thanks for review :) this bit of code is one of the things I most concerned about in the recent PRs. Previously, the "get-or-create" behaviour we had for `POST`ing to this endpoint was to look for mentees matching all of the given fields. Now, we are looking them up based on an external id (if given), otherwise using the given email address.

This means that previously, if anything changed on the matching side (e.g. first name, last name), we would be considering this a new mentee, whereas now we are only considering it a new mentee if the external id or email is different. Do you think that will be fine? Not too sure at which points on the matching platform this endpoint is hit.

codiebeulaine: @justinvdm This should be fine once we implement the external_id on the TA side. Emails are unique and so are the IDs. I'm comfortable with this but if you feel that we need more data points to match we could do that too.

PR #189, opened 2017-02-09
Add signal to delete related user when mentor or mentee is deleted

justinvdm:


codiebeulaine:
👍

PR #191, opened 2017-02-10
Show full mentor and mentee names in admin UI

justinvdm:
@codiebeulaine ready for review

PR #192, opened 2017-02-10
Use choices for call note fields instead of custom display logic

justinvdm:


justinvdm:
@codiebeulaine depends on #191, otherwise ready for review.

codiebeulaine:
👍

PR #199, opened 2017-03-02
Upgrade to django 1.10

justinvdm:
We need to do this in order to support longer usernames (there did not seem to be any alternatives out there that were easier than upgrading).

### Name conflicts issues
After upgrading from django 1.8 to django 1.10, a naming conflict became apparent: multiple inheritance was giving us Contact.mentor and Contact.mentee, and Mentor has a .mentee field. To work around this, we added a custom parent link to Mentor and Mentee so that we instead have Contact.as_mentor and Contact.as_mentee. To avoid needing to manually rename all current uses of Contact.mentor and Contact.mentee, `@property`s were created on `Contact`. This means we can access these fields as we previously did, and Mentor.mentee still refers to the relevant mentee.

**Note:** In order to change `Contact.{mentor,mentee}` to `Contact.as_{mentor,mentee}`, _modifying the migration by hand_ was necessary in order to tell django to rename the `contact_ptr` fields on `Mentor` and `Mentee`, then alter them into the new parent link fields added to the models. If we didn't do this, the `contact_ptr` fields would have been _removed_, and we would end up with `null`s where we should have foreign keys for existing rows.

### Deprecation warnings still to be resolved
```
Old-style middleware using settings.MIDDLEWARE_CLASSES is deprecated. Update your middleware and use settings.MIDDLEWARE instead.

Usage of field.rel has been deprecated. Use field.remote_field instead.

Usage of ForeignObjectRel.to attribute has been deprecated. Use the model attribute instead.

on_delete will be a required arg for ForeignKey in Django 2.0.

Specifying a namespace in django.conf.urls.include() without providing an app_name is deprecated. Set the app_name attribute in the included module, or pass a 2-tuple containing the list of patterns and app_name instead.

The app_name argument to django.conf.urls.include() is deprecated. Set the app_name in the included URLconf instead.
```

### Misc
- using lists instead of tuples in settings
- some tests needed fixing for smaller things that changed between the versions
- some deps needed to be upgraded with django, see diff of `requirements.pip`


justinvdm:

@miltontony would appreciate you reviewing too, this is a fairly dangerous change.

miltontony:
@saeedmb8 please have a read here as well

miltontony:
@justinvdm this is some amazing db mangling here!
👍 🍰 💥

justinvdm:
@miltontony thanks, and thanks for review :)

Going to land this on Tuesday, if you think that is alright. We could deploy to QA for now, but I'd rather avoid us being in a situation where deploying some minor fix before Tuesday involves us needing this to be deployed as well.

codiebeulaine:
:+1::cake:

codiebeulaine:
Post merge cake

PR #202, opened 2017-03-13
Fix feedback endpoint to handle cases where not all content is given

justinvdm:


codiebeulaine:
👍

PR #201, opened 2017-03-03
Refactoring to make way for supporting updating calls

justinvdm:

PR #203, opened 2017-03-13
Create and reconcile CDRs

justinvdm:
We need to create cdrs (call data records) given by OnionDev and reconcile cdrs given later so that we are only creating cdrs that do not yet exist.


justinvdm:

justinvdm:
Depends on #201.

justinvdm:
Ready for review now.

codiebeulaine:
👍

PR #204, opened 2017-03-22
Support OD call updates

justinvdm:
The plan is to support the current api (where call end data is given) and the new api (where updates are given). This allows us to deploy the change without worrying about backwards campatibility if OD were to rollback to the older API, and means we can (in theory) deploy the change without downtime or co-ordinating the deploy with OD (though we obviously need to test in QA with OD before deploying to production).


justinvdm:
@codiebeulaine depends #203, otherwise ready for review.

codiebeulaine:
👍 wtih

PR #209, opened 2017-03-29
Generic CSV exports

justinvdm:
At the moment, we have a download-as-csv action for messages and calls.

The action for calls is out of sync with what we display in the UI for calls.

The action for messages seems like it could be used as a generic function, but it needs some work. For one, if the client exports a csv from the admin UI, they are likely to be expecting the same field names and values as they see in the UI.

A generic action can be refactored out to solve both these problems.


codiebeulaine:
👍 🍰

PR #210, opened 2017-03-29
Show call update information in admin UI

justinvdm:
At the moment we are storing incremental call updates, we but still need to surface this information in the admin UI. This PR does that.


codiebeulaine:
👍

PR #214, opened 2017-04-05
Support excluding certain fields when downloading as csv

justinvdm:
At the moment we are exporting csvs with link fields that we don't want to export


justinvdm:

codiebeulaine:
👍 🍰 :1st_place_medal:

PR #215, opened 2017-04-13
Refactoring to make way for api v2

justinvdm:


justinvdm:
After closer inspection of the new call note designs, we can actually support new call notes without breaking backwards compatibility. Since that was the only motive for a new version of the api, I'm closing this PR for now.

justinvdm:
Ok, changed my mind, supporting new call notes without breaking backwards compatibility is proving to be more messy and complicated. Reverting back to this approach.

@nathanbegbie ready for review.

justinvdm:
:man_facepalming: @nathanbegbie I accidentally merged this instead of #218, sorry about that. I think keeping it merged and making new PRs for comments made on here might make more sense than trying to reverse merging this, thoughts?

PR #216, opened 2017-04-13
Add migrations after copy change for Call.partner_{start,end}_time

justinvdm:
Copy changes were made to display names for some fields, but the corresponding migration was forgotten.


justinvdm:

nathanbegbie:
:+1:

PR #217, opened 2017-04-18
Support v1 and v2 call notes objective achieved

justinvdm:
By supporting bool and string values, we can support both v1 and v2 of call notes for this field.


justinvdm:
Not yet ready for review unfortunately

justinvdm:
Closing in favour of the approach being started in #215.

PR #219, opened 2017-04-21
More reliable IVR-to-call linking

justinvdm:
At the moment, we link IVR feedback to calls by assuming that the latest call made to the relevant mentee is the relevant call. This is very fragile (for e.g. if we receive IVR feedback request before call request).


nathanbegbie:
@justinvdm :+1:

PR #220, opened 2017-04-21
Call Notes v2 model and api view

justinvdm:


justinvdm:
@nathanbegbie relies on #215, otherwise ready for review.

justinvdm:
@nathanbegbie Ok, ready for review again. Will add a comment for what the change in approach is for.

justinvdm:
@nathanbegbie notes on the change:

When listing call notes using `GET /v2/mentor/call_note/`, we need to show both v1 and v2 call notes. The initial plan was to just simply handle merging these lists at the view layer. This turns out to be quite difficult to do in a sane way. For example, we can't simply use django queryset pagination. Writing our own pagination component is complicated. We would need to figure out a way of listing two querysets of different models as a single list, but still having the items be ordered by some common field.

The new approach has a base call note model (which is a django-polymorphic model), and v1 and v2 models as subclasses. The base model has the common field things like `created_at` that we may want to order or filter on. This lets us list the models using django queryset pagination without needing to resort to any hackery.

Migrating existing call notes to this approach is tricky. The simplest approach I could come up with looks something like:
  - creating new models: base model, v1, v2
  - copy old call notes across to v1 (ids are preserved since v1 table is empty)
  - drop old call notes table

Both forward and backward migrations work appear to be working from the tests I've done with basic test data. Obviously, we should run this in a staging environment and see that the migration worked, and should make sure we have a dump of the db before doing this.

justinvdm:
@nathanbegbie Oh also, best to look at this one one commit at a time

mentorship/mentor/views.py
```diff
@@ -262,23 +262,34 @@ def perform_create(self, serializer):
...
     def get_queryset(self):
         contact = self.request.user.contact_details
-        return CallNote.objects.filter(mentor=contact)
+        return self.model_class.objects.filter(mentor=contact)
 
     def perform_create(self, serializer):
         serializer.save(mentor=self.request.user.contact_details)
 
 
+class CallNoteViewSet(BaseCallNoteViewSet):
```
nathanbegbie: Perhaps this should be called `CallNoteV1ViewSet`?

mentorship/mentor/serializers.py
```diff
@@ -3,9 +3,10 @@
 from django.contrib.auth.models import User
 
 from mentor.utils import get_image_url, normalise_contact_number
-from mentor.models import Call, Message, Event, Task, ScheduledCall, Mentee, \
-    Mentor, NotificationSettings, Category, Activity, CallNote, Topic, FAQ, \
-    Blog, Post, PostLike
+from mentor.models import (
+    Call, Message, Event, Task, ScheduledCall, Mentee, Mentor,
+    NotificationSettings, Category, Activity, CallNote, CallNoteV2, Topic, FAQ,
```
nathanbegbie: Should we not be importing CallNoteV1 here?

mentorship/mentor/views.py
justinvdm: :+1:

nathanbegbie:
Minor comments, but :+1:

PR #224, opened 2017-05-02
Add a version attribute to call notes

justinvdm:
We need this for api clients to differentiate between call note versions when listing call notes.

@nathanebegbie ready for review


nathanbegbie:
:+1:

PR #228, opened 2017-05-10
Feature/return most recent calls for mentor

nathanbegbie:


nathanbegbie:
@justinvdm please review

mentorship/mentor/tests/test_views.py
```diff
@@ -926,6 +926,16 @@ def test_calls_list(self):
...
 
+        # test that only most recent 100 calls are returned
+        calls = [self.mk_call(mentor=mentor) for i in range(101)]
+        data = [CallSerializer(call).data for call in calls]
+
+        response = self.client.get(url)
+
+        results_calls = json.loads(response.content)['results']
+        self.assertEquals(len(results_calls), 100)
+        self.assertEquals(data[-100:][::-1], results_calls)
```
justinvdm: I think it would be enough to just create two calls with their start times set explicitly, and see that they are ordered in descending order. Thoughts?

justinvdm:
One minor comment, looks good :) I think travis is failing because of a linting error.

mentorship/mentor/tests/test_views.py
nathanbegbie: Yeah, I realise that creating 101 calls is not very pretty, but it occurred to me that what this does is explicitly document that we expect only the most recent 100 calls from this api.
Happy to change this though.

justinvdm:
Post-merge :+1:

PR #231, opened 2017-05-23
blank=True for some call note v2 fields

justinvdm:
Some steps are skipped in certain cases, these fields should be `blank=True`.


nathanbegbie:
@justinvdm Could we provide better naming for migrations? Otherwise +1

justinvdm:
@nathanbegbie landing given our discussion earlier on slack about commit ammending

